Issue #30, opened 2015-11-03
Clean up checkstyle


phillipross:
The existing checkstyle setup should be cleaned up and more standardized.  The rules should be replaced with more standard conventions as a base (sun_checks) and progressively tweaked with project-specific customizations.

phillipross:
Merged PR #32 to begin this work.

phillipross:
Changes to manta client and httpsigner are completed, but not reviewed yet.
Changes to checkstyle.xml have been made, but config package still needs to be updated.

dekobon:
Looks good to me. :+1: 

What's Puppy Crawl?

phillipross:
I guess Puppy Crawl is/are/was/were the developer(s) of the actual checkstyle component.

I had to bump up the version of that component since the maven plugin was referring to an older version which didn't know to skip over serialVersionUID fields when looking for non-javadoced members of a class.

Thanks for looking at these changes... I'll merge PR #36  into the checkstyle branch.  If you're doing your checkstyle work directly on the joyent repo checkstyle branch, please be sure to pull/merge to get the latest updates from PR #36

PR #32, opened 2015-11-03
Initial checkstyle configuration

phillipross:
Configured initial set of checkstyle rules, but keep the build configured to skip the checkstyle checks by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ which can be configured
 
 # Contributions
 
-Contributions welcome! Please ensure that `# mvn checkstyle:checkstyle` runs
+Contributions welcome! Please ensure that `# mvn checkstyle:checkstyle -Dcheckstyle.skip=false` runs
 clean with no warnings or errors.
 
 ## Testing
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,96 +1,157 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
 
-<!-- This configuration file was written by the eclipse-cs plugin configuration
-  editor -->
-<!-- Checkstyle-Configuration: java-manta Description: Checkstyle configuration
-  that checks the sun coding conventions. -->
+<!--
+  Checkstyle configuration that checks the sun coding conventions from:
+    - the Java Language Specification at
+      http://java.sun.com/docs/books/jls/second_edition/html/index.html
+    - the Sun Code Conventions at http://java.sun.com/docs/codeconv/
+    - the Javadoc guidelines at
+      http://java.sun.com/j2se/javadoc/writingdoccomments/index.html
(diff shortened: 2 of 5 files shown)


phillipross:
@dekobon If you get a free moment, can you take a quick look at these changes to fixup the checkstyle integration for the build?

It's a baseline configuration based on the sun conventions.  I put documentation into the configuration with references to the checkstyle documentation which contains rationales for the various rules as well as configuration options for the various rules.  By default, the checkstyle validation is skipped, but can be set with a system property (see the README.md change).

As it stands now... even with some rules commented out, there are about 70 checkstyle violations (:laughing:) of which most of them are simply things like member names ending with underscores or not having the correct capitalization schemes, as well as missing curly braces for if statements.  We can discuss the various checks and decide which to keep and which to throw away, then fix the violations going forward.

Thanks!

dekobon:
I just read over it. The change looks good to me. :+1: 

What do you think of fixing the style violations in one go? We can tackle them in a branch and split the work 50/50? Also what do you think of the javadoc formatting? Something about it seems odd.

phillipross:
Yeah, I'm all for fixing the violations now.  I'll merge this one to master and create a separate branch for the violation fix work.

There are actually a few rules that are still commented out... like the javadoc ones, design for extension, hide utility class constructors, et al.  Some of these I'm not sure if they should be uncommented or not.  I wouldn't necessarily know how to handle some of them, like the "missing switch default"

Some of them might be contentious as well.  For example, I'm not fond of the "avoid star imports" but I understand the rationale and have seen plenty of bugs due to the use of star imports... but many people push back hard and want their star imports.

The other big one is the "avoid inline conditionals" that seems to upset many.  I hate inline conditionals in almost all cases, but many people swear that it makes the code more concise.  But in all my years of experience, they are the source of many bugs due to readability problems... right up there with regular expressions.

Anyway, we can talk about specific ones as they come up.

And you're right about the existing javadoc.  I noticed there are a lot of newlines or spacing weirdness where there doesn't seem to be an obvious need.  Feel free to fix any of that stuff where ever you see it, I don't mind :smirk:

dekobon:
I'm actually ok with a some-what liberal checkstyle. I used to be a style
nazi and then I just ended up having too much friction on teams, to I just
gave up.

Let's just uncomment them all and see what fails, if it is a stupid or
burdensome rule, then let's recomment it.

I see star imports as fine. Inline conditionals - it depends. If it is the
return statement on a one line method, why not?

On Tue, Nov 3, 2015 at 1:21 PM, Phillip Ross <notifications@github.com>
wrote:

> Merged #32 <https://github.com/joyent/java-manta/pull/32>.
>
> —
> <https://github.com/joyent/java-manta/pull/32#event-453712473>.
>



-- 
-Elijah

phillipross:
Hah, I was the same way... but there's a point of diminishing returns on spending time debating conventions versus the time spent fixing bugs that could have been avoided by said conventions. 

I'm all for uncommenting all of them and then working from there :+1:

I already started working on MantaClient.java to see what I'd run into... I'll work on the `checkstyle` branch and reference issue #30 in the commit messages so they show up under the issue in github.

PR #37, opened 2015-11-04
Checkstyle

dekobon:
@phillipross I finished up the checkstyle changes. I did end up commenting out two checkstyle checks. Please review.


phillipross:
@dekobon I figured you'd want to re-comment out the DesignForExtension rule as it enforces some fairly rigid OOP design fundamentals with respect to how one defines and implements class hierarchies with open extensibility AND maintaining strict encapsulation at the same time.  :grin: 

But it looks like you also commented out the hidden field rule, but I see now that I forgot to add the option for builder/fluid setter config to allow the hidden field rule to properly recognize those types of settings.  Can you do another commit that uncomments the rule, but adds this element as an option to the rule:
```
<property name="setterCanReturnItsClass" value="true"/>
```

If this doesn't work and it still complains about your setters hiding the fields, comment the rule back out and I'll take a look at the checkstyle configs at some point to see what the expression matcher logic is.

PR #50, opened 2015-11-10
Refactored project to use Apache HTTP Client to back all HTTP requests

### LINKED ISSUE #31: Factor out HttpSigner into separate repo
A [new repo](https://github.com/joyent/java-http-signer) has been setup for this.

It would probably be better to cleanup the build/release/document/conventions here before factoring this out into the separate repo.  Dependent issues include #29 & #30

### LINKED ISSUE #44: Change HTTP library to Apache HTTPClient
By making this change we will be able to get a more feature rich library that will allow us to do things like Range requests and use pools.

https://hc.apache.org/httpcomponents-client-ga/

dekobon:
Fixes #31 
Fixes #44 

@phillipross Please review when you get a moment.


phillipross:
@dekobon This looks great.  But I see that the 1.0.0 version of the signature dependency is referenced here.  You released the 1.0.0 version of the signature artifact, but there is some lagtime between the time you release via sonatype and the time when it is available in maven central.  Could you wait until the signature artifact appears in maven central before merging this branch?  Thanks!

dekobon:
Of course, I forgot about that!

phillipross:
Looks like the artifacts are in central now.  It's OK to merge this PR :+1:

Issue #48, opened 2015-11-10
README specifies incorrect system property name for manta user


phillipross:
`manta.url` should be `manta.user` in the system property column for the user property

Issue #53, opened 2015-11-11
isDirectory() doesn't work when you do MantaClient.get() for a directory


dekobon:

dekobon:
@phillipross I've temporary broken the build by adding a failing test for this issue.

PR #55, opened 2015-11-11
Bugfixes

phillipross:
Fixes https://github.com/joyent/java-manta/issues/48 https://github.com/joyent/java-manta/issues/53 and https://github.com/joyent/java-manta/issues/56
Also tweaks logback for tests and gets dependency-reduced-pom.xml properly untracked in git.


dekobon:
Looks good to me. :+1:

PR #64, opened 2015-11-15
Stream rejigger

### LINKED ISSUE #40: Refactor how streams are handled

dekobon:
Fixes #40 

@phillipross I know this is a huge PR, but could you review at your convenience?


phillipross:
No problem, I'll review it... i've been watching the branch to see how things are progressing.

src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -273,24 +226,24 @@ public MantaClient(final String url,
...
-        // XXX: make an intercepter that sets these timeouts before each API call.
-        request.setReadTimeout(this.httpTimeout);
-        request.setConnectTimeout(this.httpTimeout);
 
         HttpResponse response = null;
         try {
             response = request.execute();
-            LOG.debug(String.format("got response code %s, header %s ",
-                                    response.getStatusCode(), response.getHeaders()));
+            if (LOG.isDebugEnabled()) {
```
phillipross: Are these isDebugEnabled checks necessary?

phillipross:
This all looks good, nice work!  I'm going to push a snapshot build to sonatype staging repo (not for maven central) and kick around the snapshot with an external project before we merge.  I'll follow up here with any comments.

phillipross:
OK, two things that are still problematic for me.

1. In order for client code to NOT have to be coupled with google http client and be able to do payload integrity checking, the following must be done:

 ```java
mantaClient.put(
    PATH_PREFIX + assetFilename,
    new ByteArrayInputStream(assetByteArray),
    (new MantaObjectMetadata(""))
                   .getHttpHeaders()
                   .set("Content-MD5", Arrays.asList(base64EncodedMD5Digest))
);
```
 In this snippet, the put method takes a headers param that is an an implementation from the google http client package.  Client code can get around tight coupling with google http client by using a pre-existing instance of the headers class, and this code snippet shows instantiation of `MantaObjectMetadata` and using it's headers instance, but an empty path is required as a parameter to the constructor since the headers member is not instantiated in the default no-args constructor.
 It would be better for java-manta to have some abstraction over google http client api so that clients do not have to be bound to google's http client api.

2. I'm getting timeouts waiting for connections from the http connection pool.  In 1.x codeline, I explicitly had to close the dataInputStream on all instances of `MantaObject`, but with this new codeline, that input steam no longer exists.  It would be a good idea to have a test method in the test suite which performs a lot of operation in a row in an attempt to detect leaks.

File: src/main/java/com/joyent/manta/client/MantaClient.java [hunk 2]
```diff
@@ -338,81 +290,221 @@ public void deleteRecursive(final String path) throws MantaCryptoException, Mant
...
+     * @return The {@link MantaObjectMetadata}.
+     * @throws IOException If an IO exception has occurred.
+     * @throws com.joyent.manta.exception.MantaCryptoException If there's an exception while signing the request.
      * @throws MantaClientHttpResponseException If a http status code {@literal > 300} is returned.
      */
-    public MantaObject get(final String path) throws MantaCryptoException, MantaClientHttpResponseException,
-    IOException {
-        LOG.debug(String.format("entering get with path %s", path));
+    public MantaObjectMetadata get(final String path) throws IOException {
+        final HttpResponse response = httpGet(path);
```
phillipross: Does this `HttpResponse` instance need to be explicitly disconnected?  I noticed the `httpPut` method contains a finally block which explicitly invokes the disconnect method of the response, but I don't see anything explicit here or in any of the invokers of this method.  Could this be causing the leak that I'm seeing with the http client's connection pool?

src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: They aren't I think they were in my first version and then it hung around as cruft.

dekobon: Oh, I remember why... We parameterize method calls (response.getStatusCode(), etc), so I have it ingrained in me as a habit to do a isXXEnabled() check around it when I am not doing simple interpolation with fields. Since, these are getters, it shouldn't matter.

src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: Yes, it does. Nice catch. There is a bunch of logic in Apache HTTP Client to deal with stale connections in a pool, so in theory leaving these open isn't a big deal, but we all know about the different between theory and practice... Will be changed in the next update.

phillipross:
commit b670c02 addressed the problem with exhausting the http connection pool.
As of f1a595a this branch is very usable for me.

dekobon:
That's great news. I'm busy working on metadata support right now. There's a wart in how the Google HTTP client works. It forces you to set a Content-Length header. This makes the Manta barf, so I'm implementing the request using the raw Apache HTTP client.

phillipross:
Does it force you to set the content-length if you set the encoding to chunked?

dekobon:
That seemed to do the trick. Here's a question for you. How are you reading metadata? After I PUT it, I can't seem to get a way to read it.

phillipross:
When you say "metadata" are you referring to the Manta service's object metadata that is access via the "m- headers"  ?

Currently, when I do a PUT, I have to specify that metadata by getting an instance of `com.google.api.client.http.HttpHeaders` and populate headers with the metadata.

In the 1.x codeline the MantaClient class has `put(MantaObject mantaObject)` method, and that MantaObject instance encapsulates the http headers.  In the 2.x codeline w/ stream_rejigger branch, the put method accepts the path and http headers directly instead of using a MantaObject.  In this case, I have to either build my own MantaObject from the values I used to invoke the PUT call... or do a GET after the PUT to get a populated instance back from the Manta service.

I was waiting to see how the first merge-able cut of the code shook out before filing any issues against it, but I wanted a more streamlined way to do the above... maybe having the put method return a populated manta object instance or something.

src/main/java/com/joyent/manta/client/MantaHttpHeaders.java
```diff
@@ -0,0 +1,161 @@
...
+import static com.joyent.http.signature.HttpSignerUtils.X_REQUEST_ID_HEADER;
+
+/**
+ * Object encapsulating the HTTP headers to be sent to the Manta API.
+ * When non-standard HTTP headers are used as part of a PUT request to
+ * Manta, they are stored as metadata about an object.
+ *
+ * @author <a href="https://github.com/dekobon">Elijah Zupancic</a>
+ */
+public class MantaHttpHeaders extends HttpHeaders {
```
phillipross: The problem with having this inherit from HttpHeaders is that HttpHeaders contains a lot of methods which return itself so it can be manipulated fluently.  These methods return `com.google.api.client.http.HttpHeaders` instead of `com.joyent.manta.client. MantaHttpHeaders` which messes up chained invocation :disappointed:

dekobon:
For some reason, when reading the Manta docs, I didn't understand that I needed to prefix all Manta metadata with "m-" - hence my question on how you were retrieving it.

As for encapsulating it so that it doesn't expose the underlying HttpHeaders object, that was easy enough to do by change from inheritance to a delegation model.

Lastly, I'm trying to find the best way to expose metadata creation and access. Right now, we are conflating the word metadata all over the place. We have a `MantaMetadataObject` we have HTTP headers, but we have an accessor that allows you to get the metadata from the headers.

I'm thinking that I rename `MantaMedataObject` to `MantaResultObject` and create a method that allows for access to metadata separate from the headers on the `MantaObject` interface.

Also, I'm not sure if I should insulate the user of the library from the `m-` prefixing when they are using the metadata specific API.

What are your thoughts?

phillipross:
I agree... the word "metadata" makes it challenging to understand the semantics of many things.  I haven't really even looked at the manta documentation deep enough to know whether the letter **m** is short for "manta" or "metadata"

It's probably safest to NOT insulate the consumers of the library from prefixing with "m-" at first.  As long as the user is dealing with an **http header** I think they should have to deal with the actually names of those headers... the raw names.  If we completely abstract headers and have things like durability and content-md5 as properties of a manta object, then maybe the same abstraction can be applied to "m-" headers as well.  

I hope that makes sense :grinning:

PR #68, opened 2015-11-18
Added javadoc indicating that we only support a single metadata value per key.

### LINKED ISSUE #66: http headers should be case-insensitive
According to the [RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

It would be a good idea to verify that the manta service also behaves in a way that treats the headers as case-insensitive, then validate that java-manta evaluation of header/metadata key names is also case-insensitive, especially when dealing with "m-" custom metadata.

dekobon:
Fixes #66
@phillipross What do you think?


phillipross:
perfect :+1:

PR #69, opened 2015-11-19
Add home directory to context and fixed unit tests to support sub users

dekobon:
@phillipross please review if you have a free moment.

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/config/ConfigContext.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/config/ConfigContext.java
@@ -34,4 +34,33 @@ public interface ConfigContext {
      * @return General connection timeout for the Manta service.
      */
     Integer getTimeout();
+
+    /**
+     * @return String of home directory based on Manta username.
+     */
+    String getMantaHomeDirectory();
+
+    /**
+     * Extracts the home directory based on the Manta account name.
+     *
+     * @param mantaUser user associated with account
+     * @return the root account holder name prefixed with a slash or null
+     *         upon null mantaUser
+     */
+    static String deriveHomeDirectoryFromUser(final String mantaUser) {
+        if (mantaUser == null) {
+            return null;
+        }
+
+        final String home;
+
+        if (mantaUser.contains("/")) {
+            final String accountUser = mantaUser.split("/")[0];
+            home = String.format("/%s", accountUser);
+        } else {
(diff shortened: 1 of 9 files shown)


phillipross:
I took a look at the code, and am I correct in my interpretation that you're using some heuristic that defines a user to be considered the owner of the account if the user begins with a "/" but not necessarily the account holder otherwise?

dekobon:
Yes, if the username contains a "/", then it is a subuser account because slash is a reserved character.

phillipross:
Perhaps you mean the account owner begins with a slash?  As long as it's eventually documented in the readme or something, I think any potential confusion will be minimal :grinning:

dekobon:
Yes. That's what I meant. Yeah, I wonder where I will document this.

On Wed, Nov 18, 2015 at 6:03 PM, Phillip Ross <notifications@github.com>
wrote:

> Perhaps you mean the account owner begins with a slash? As long as it's
> eventually documented in the readme or something, I think any potential
> confusion will be minimal [image: :grinning:]
>
> —
> <https://github.com/joyent/java-manta/pull/69#issuecomment-157923659>.
>



-- 
-Elijah

phillipross:
I would assume a section in the README.md would suffice for now.  There's no wiki as of yet, but I dont know if there are internal standards that joyent as that should be adhered to or what.

dekobon:
@phillipross please have a look at the doc changes when you have a moment.

java-manta-client/src/main/java/com/joyent/manta/config/ConfigContext.java
```diff
@@ -34,4 +34,33 @@
...
+     *         upon null mantaUser
+     */
+    static String deriveHomeDirectoryFromUser(final String mantaUser) {
+        if (mantaUser == null) {
+            return null;
+        }
+
+        final String home;
+
+        if (mantaUser.contains("/")) {
```
phillipross: I'd recommend a slightly stronger validation here... maybe check to make sure the manta user config value contains 0 or 1 "/" characters but no more.  And also make sure the "/" is not at the beginning, or the 0-th element is going to result in an empty string (I think).

phillipross:
After seeing the documentation, I understand how it works now... I was confused by that slash character.  I annotated the `deriveHomeDirectoryFromUser(...)` method with a comment about validation, but I'd say this is good to push as-is.  :+1:

PR #71, opened 2015-11-20
Added better testing around subusers.

dekobon:
I also added dependencies on test classes, so that we can't create a directory, the majority of other tests won't run.


dekobon:
@phillipross please review when you have a moment.

phillipross:
:+1:

PR #75, opened 2015-11-21
Fixes #61

### LINKED ISSUE #61: MantaClient should interpret JSON content for manta errors and build structured exceptions
Currently, various MantaClient methods throw MantaClientHttpResponseException which does not provide a sufficient interface which applications can use to understand the underlying error.  The manta service sends back JSON content describing the error, but the manta client builds the exception based on the http-oriented semantics (http status code) rather than the manta-oriented semantics (manta service error code).

Applications ARE able to access the JSON that manta sends back, but it's up to the application to parse the JSON and interpret the error content.

dekobon:
Adds error response enum. Parses JSON error response from Manta API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ matrix:
         - DESC="oraclejdk8 unit tests"
         - CMD="mvn clean verify -DskipITs=true"
 
-script: ${CMD}
+script: echo ${CMD}; ${CMD}
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,5 @@ matrix:
       env:
         - DESC="oraclejdk8 unit tests"
         - CMD="mvn clean verify -DskipITs=true"
+
+script: ${CMD}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added build to Travis CI.
 - Added closeQuietly() to MantaClient.
 - Added parsing of JSON error responses from Manta API.
+- Added error response enum MantaErrorCode.
 
 
 ## [1.6.0] - 2015-11-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added MantaMetadata class for handling object metadata.
 - Added MantaHttpHeaders class for handling custom HTTP headers.
 - Added build to Travis CI.
-- Added closeQuietly() to MantaClient
+- Added closeQuietly() to MantaClient.
+- Added parsing of JSON error responses from Manta API.
 
 
 ## [1.6.0] - 2015-11-11
--- a/java-manta-client/src/main/java/com/joyent/manta/client/MantaUtils.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/MantaUtils.java
@@ -210,16 +210,16 @@ public final class MantaUtils {
         if (account.isEmpty()) {
             throw new IllegalArgumentException("Username can't be empty");
         } else if (slashPos == -1) {
-            return new String[] { account };
+            return new String[] {account};
         } else if (slashPos == 0) {
             throw new IllegalArgumentException("Username can't begin with /");
-        } else if (account.charAt(account.length() -1) == '/') {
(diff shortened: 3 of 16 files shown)


dekobon:
@phillipross Please review when you have a free moment.

phillipross:
The code looks good, but for some reason, when I've checked out the branches the last few times to browse the code within the IDE, the tests don't seem to run.  I wasn't sure if you had temporarily disabled these while you were working on getting the CI up and running or if this was a mistake or something.  I think there might be a little more work necessary for the test suite wiring.

Also... the CI doesn't seem to be checking the checkstyle rules.  Is that on purpose as well?

phillipross:
Oops, I think I see what the problem is with the testsuite.  I can fix it but I'm not sure what your intentions were with how the CI runs which test suites and when.  I would assume the we'd want the CI to be able to invoke a build with and without style checks and tests suite checks... and this should be able to be driven with -D system properties on the maven invocation.  Is that correct?

dekobon:
@phillipross, 3f0d373dc3b9dc568e85c3c5208945fbff9986f0 should resolve the issues with tests and style checks running in CI. As for your IDE, I'm not sure what the problem is. Could it be that I added some dependsOn annotations to TestNG? All of the tests are running on my side. I'm going to go ahead and merge.

phillipross:
I have no problems specifically with the IDE, I was just referring to the process of reviewing the PRs by looking at the diffs rendered by github versus checking out the code into my IDE and reviewing it in the IDE editor.

The checkstyle problem seems to stem from the fact that the default property value for checkstyle is set to skip the checkstyle checks: `<checkstyle.skip>true</checkstyle.skip> `

The travis config does not seem to explicitly override this.  It needs `-Dcheckstyle.skip=false`

Also, I dont think the `-DskipITs=true` is doing anything.  The integration tests don't run due to some maven configuration.  I'm actually traveling at the moment, but I'll be back at my desk tomorrow and can provide more info on the maven config.  Thanks!

dekobon:
In the .travis.yml, I specified:

```yaml
language: java
sudo: false

before_install:
  - echo 'MAVEN_OPTS="-Dorg.slf4j.simpleLogger.defaultLogLevel=warn"' >~/.mavenrc

matrix:
  fast_finish: true
  include:
    # checkstyle
    - jdk: oraclejdk8
      env:
        - DESC="checkstyle"
        - CMD="mvn clean checkstyle:checkstyle -Dcheckstyle.skip=false"
    # unit tests (oraclejdk8)
    - jdk: oraclejdk8
      env:
        - DESC="oraclejdk8 unit tests"
        - CMD="mvn clean verify -DskipITs=true"

script: echo ${CMD}; ${CMD}
```
Do you think that this is broken somehow?

phillipross:
Woops, I didn't see the dedicated checkstyle rule there.
I see it now, and it looks correct, but I know the build was passing even when there were checkstyle errors.  I'll take a closer look at the current state though.

phillipross:
I see now... the checkstyle invocation as you have it does not work, but if you change it to use the validate phase which is how the checkstyle plugin is bound into the build, it should work:
`mvn clean validate -Dcheckstyle.skip=false`

I'm not immediately certain why it doesn't work when invoking the checkstyle plugin directly, but I can look into it if you want.

phillipross:
Hah, ok ok... the `checkstyle` goal is a reporting goal for generating reports.  The `check` goal is the one for immediately running checkstyle.  The `check` goal is the one that is bound to the `validate` phase.

phillipross:
I'm about to submit a PR that will refactor the maven configs for testing to enable the integration tests execution.  I'll include changes to the travis config file to modify the checkstyle rule (I want to test travis behavior with PRs) so you may see some weird initial commits on the PR.  Please don't try to merge the PR before I get the real changes in :grinning:  thanks!

dekobon:
Ok. Sounds great!

On Mon, Nov 23, 2015 at 2:38 PM, Phillip Ross <notifications@github.com>
wrote:

> I'm about to submit a PR that will refactor the maven configs for testing
> to enable the integration tests execution. I'll include changes to the
> travis config file to modify the checkstyle rule (I want to test travis
> behavior with PRs) so you may see some weird initial commits on the PR.
> Please don't try to merge the PR before I get the real changes in [image:
> :grinning:] thanks!
>
> —
> <https://github.com/joyent/java-manta/pull/75#issuecomment-159090095>.
>



-- 
-Elijah

PR #78, opened 2015-11-23
maven test plugin configs

phillipross:
Slight refactorings for maven test plugin configs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     - jdk: oraclejdk8
       env:
         - DESC="checkstyle"
-        - CMD="mvn clean checkstyle:checkstyle -Dcheckstyle.skip=false"
+        - CMD="mvn clean checkstyle:check -Dcheckstyle.skip=false"
     # unit tests (oraclejdk8)
     - jdk: oraclejdk8
       env:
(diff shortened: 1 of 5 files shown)


phillipross:
@dekobon From my perspective, this PR is ready for merging.  There are a few of the commits w/ travis failures, and those were all expected... so I now have observed travis functioning correctly :grinning: 

This is an explanation of what I did if the diffs are not explanatory enough:
- I pushed the surefire configs down into manta-client module.  The manta-client module contains the "unit tests" which are invoked by surefire during the `test` phase.
- I pushed the failsafe configs down into manta-client-it module.  The manta-client-it module contains the "integration tests" which are invoked by failsafe during the `integration-test` phase, and then verified to have passed or failed during the `verify` phase.
- I enabled the checkstyle plugin to run by default by changing the `checkstyle.skip` property from true to false.
- I changed the travis config to use the correct maven invocations for checkstyle check and unit test check.  Since only unit tests are run for now... only the test lifecycle needs to be invoked and not the verify lifecycle.

By default, checkstyle check, unit test, and integration tests are all enabled by default.  When future contributors begin to contribute changes, we'll want to motivate them to be sure that checkstyle rules are not violated, unit tests don't break, and integration tests don't break... so it makes most sense to keep these all enabled by default.  Each one of these can be skipped by specifying their respective system property.

dekobon:
This looks solid. Let's check it in.

PR #79, opened 2015-11-24
Adds msign like support. You can HTTP sign Manta URLs.

### LINKED ISSUE #73: Add msign support

dekobon:
Fixes #73 

@phillipross please review at your leisure.

### DIFF
--- a/java-manta-client/src/test/java/com/joyent/manta/client/MantaHttpHeadersTest.java
+++ b/java-manta-client/src/test/java/com/joyent/manta/client/MantaHttpHeadersTest.java
@@ -1,7 +1,7 @@
 package com.joyent.manta.client;
 
-import junit.framework.Assert;
 import org.apache.commons.collections4.CollectionUtils;
+import org.testng.Assert;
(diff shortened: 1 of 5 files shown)


dekobon:
For commit "Minor pom.xml cleanup." I was running on auto-pilot and gave it a very bad message.

There were two changes:
1. Removed com.joyent form the list of shade exclusions because it was doing nothing.
2. I added a filter to explicitly exclude the class: org/apache/commons/collections4/functors/InvokerTransformer.class when shading because this class is of concern for a number of the Java serialization vulnerabilities.

java-manta-client/src/test/java/com/joyent/manta/client/MantaHttpHeadersTest.java
```diff
@@ -0,0 +1,33 @@
+package com.joyent.manta.client;
+
+import junit.framework.Assert;
```
phillipross: Oops, I'm assuming this is an incorrect import.  should be testng package instead of junit

PR #85, opened 2015-11-25
Fixes #17

### LINKED ISSUE #17: Normalize all parameters and variables that refer to Manta account with different terminology

dekobon:
@phillipross Please review when you have a moment.

The key part of this change is reflected in the readme:

Joyent's SmartDataCenter account implementation is such that you can have a
subuser as a dependency upon a user. This is part of SmartDataCenter's [RBAC
implementation](https://docs.joyent.com/public-cloud/rbac/users). A subuser
is a user with a unique username that is joined with the account holder's 
username. Typically, this is in the format of "user/subuser".
 
Within the Java Manta library, we refer to the account name as the entire
string used to login - "user/subuser". When we use the term user it is in
reference to the "user" portion of the account name and when we use the term
subuser, it is in reference to the subuser portion of the account name.

The notable exception is that in the configuration passed into the library,
we have continued to use the terminology *Manta user* to refer to the
account name because of historic compatibility concerns.

I did my best to change how we used the terms so that they were consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ left are overridden by values on the right.
 The URL of the manta service endpoint to test against
 * `manta.user` ( **MANTA_USER** )
 The account name used to access the manta service. If accessing via a [subuser](https://docs.joyent.com/public-cloud/rbac/users),
-you will specify the username as "account/subuser".
+you will specify the account name as "user/subuser".
 * `manta.key_id`: ( **MANTA_KEY_ID**)
 The fingerprint for the public key used to access the manta service.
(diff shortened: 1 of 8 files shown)


phillipross:
Code looks fine, and the text is at least more consistent within java-manta :+1: 
I may be experiencing a little cognitive dissonance at this point, but I think we'll have to iterate over it and handle inconsistencies w/ [java-http-signature](https://github.com/joyent/java-http-signature) on some future passes :grinning:

PR #104, opened 2016-01-07
Integration with the latest HTTP Signature library

dekobon:


phillipross:
These changes look fine too, but it looks like we have a workflow issue with CI.
Maybe the maven build needs to be changed to specify snapshots (and maybe releases too?) be pulled from sonatype OSS repos so that snapshot builds can be published and dependent projects can resolve the artifacts.  This can be tackled separately though.

PR #120, opened 2016-10-24
Adds Multipart upload functionality implemented in jobs

dekobon:


geek:
LGTM

I confirmed with the following app:


```java
package com.joyent.app;

import com.joyent.manta.client.MantaClient;
import com.joyent.manta.client.MantaMultipartManager;
import com.joyent.manta.client.MantaMultipartUpload;
import com.joyent.manta.config.ChainedConfigContext;
import com.joyent.manta.config.ConfigContext;
import com.joyent.manta.config.DefaultsConfigContext;
import com.joyent.manta.config.StandardConfigContext;

import java.io.IOException;

public class App {
    public static void main(String... args) throws IOException {
        ConfigContext defaultConfig = new DefaultsConfigContext();
        ConfigContext customConfig = new StandardConfigContext()
                .setMantaKeyId("PUBLIC KEY MD5 ID")
                .setMantaUser("MY USERNAME");
        ConfigContext config = new ChainedConfigContext(defaultConfig, customConfig);

        MantaClient client = new MantaClient(config);

        MantaMultipartManager multipart = new MantaMultipartManager(client);

        MantaMultipartUpload upload = multipart.initiateUpload("~~/stor/file");
        System.out.println(upload.getId());
        System.out.println(upload.getPath());

        // We can add the parts in any order
        multipart.putPart(upload, 2, "Multipart\n");
        // Each put of a part is a synchronous operation
        multipart.putPart(upload, 1, "Hello From\n");
        // Although in a later version we could make an async option
        multipart.putPart(upload, 3, "File Upload\n\n");


        // We've uploaded all of the parts, now lets join them
        multipart.complete(upload);
        multipart.waitForCompletion(upload);
    }
}
```

```sh
$ mget ~~/stor/file
Hello From
Multipart
File Upload


```

cburroughs:
I was also able to run a trivial example.  I think my broadest feedback would be to consider having a `MantaMultipartManager` interface implemented by  `MantaMultipartCllientJobManager` (or whatever) so there is a clear way to incorporate server-side support, and we are sure the  java api makes sense.  Unless the intent is *not* to have the same java api with server side support, in which case this needs some "Here Be Dragons" warnings to avoid wayward users against relying on it.

 * When I used bogus credentials I got`{"code":"InternalError","message":"an unexpected error occurred"}`, but presumably we know that's some sort of auth error?
 * I'm not sure why all the upload related classes need to be `Serializable`.
 * I  think you could make the private fields of `MantaMultipartUpload` final, since there are no setters.
 * Is `MULTIPART_DIRECTORY` a random UUID? If so I think there should be a comment to that effect.
 * Why does `MantaMultipartUpload` need to be comparable?
 * I tried aborting an upload and got an NPE at `MantaMultipartManager.java:587`
 * Should there be some sort of assertion on `Part.partNumber` being positive?
 * If we are already depending on a library that can generate them, a time based UUID might help debug some odd corner case where `mtime` is insufficient down the road.
 * In `MantaMultipartManager.listInProgress` there is a list of Throwables, but I *think* from the code only Exceptions could end up there.  (Catching Throwables tends to be scary.).  Also, I'm unsure if it is correct that `count` in that same method starts at 1.
 * I'm not sure `Part` benefits from being an inner class in the long run. (Again I'm unsure of the intended longevity of this Java API.)
 * I suspect `validateThereAreNoMissingParts` is going to be hard to use as written.  `MantaClientException` wouldn't be what I would initially expect for missing parts
 * In the spirit of Joyent conventions, should `MultipartMetadata` be serialized as JSON?
 * In `waitForCompletion`, what happens if we exceed `timesToPoll`?
 * I know this is a tide-over POC, but if I `mrm` one of the parts mid upload, the commit still succeeded. 
 * Tried to upload 9000 files, but the commit failed with `{"code":"RequestEntityTooLargeError","message":"Request body size exceeds 100000"}`  I don't think the string builder can handle near 10k (maybe that's okay initially?)
 * I'm worried that the current semantics of committing ("cat everything in the upload dir together") are significantly different from what we want the final API to be ("commit *these* etags").

dekobon:
@cburroughs, Please see my replies below.

> I was also able to run a trivial example. I think my broadest feedback would be to consider having a MantaMultipartManager interface implemented by MantaMultipartCllientJobManager (or whatever) so there is a clear way to incorporate server-side support, and we are sure the java api makes sense.

My intention was to throw away the jobs based implementation entirely once we add support on the server. This would allow for little to no code changes on the part of the consumer of the Java SDK. Whereas if we did the interface to multiple implementation method, the consumer would have to change their code.

> When I used bogus credentials I got{"code":"InternalError","message":"an unexpected error occurred"}, but presumably we know that's some sort of auth error?

This is probably a deficiency in the SDK itself. Also, that error isn't typical for an authentication failed error. Can you do a PR on this branch adding an integration tests that causes that message to be produced?

> I'm not sure why all the upload related classes need to be Serializable.

They don't need to be serializable once we have support at the server. However, I do use Java serialization to store state on Manta for the jobs-based implementation. I don't think it hurts anything. Also, there is a cult among the larger Java religion that believes that all classes should be marked as Serializable unless there are truly intended not to be. That is why classes like Exception implement Serializable - I mean who wants to serialize exceptions?!? I have no stake in this game.

> I think you could make the private fields of MantaMultipartUpload final, since there are no setters.

Done.

> Is MULTIPART_DIRECTORY a random UUID? If so I think there should be a comment to that effect.

It is a randomly chosen UUID that is the same for all users. I just wanted a value that wasn't likely to conflict with any other directory names. Comment added.

> Why does MantaMultipartUpload need to be comparable?

It doesn't need to, but old habits die hard. I just assumed that you would want a default sort order by path names if you executed .sorted() on a stream of multipart listings.

> I tried aborting an upload and got an NPE at MantaMultipartManager.java:587

I will look into this and update the PR with my findings.

> Should there be some sort of assertion on Part.partNumber being positive?

There is already a check within the [MantaMultipartUploadManager](https://github.com/joyent/java-manta/blob/multipart/java-manta-client/src/main/java/com/joyent/manta/client/multipart/MantaMultipartManager.java#L817). However, I did go ahead and add an additional check in the MantaMultipartUploadPart object itself.

> If we are already depending on a library that can generate them, a time based UUID might help debug some odd corner case where mtime is insufficient down the road.

I don't quite understand what you mean. Could you explain this further?

> In MantaMultipartManager.listInProgress there is a list of Throwables, but I think from the code only Exceptions could end up there. (Catching Throwables tends to be scary.). Also, I'm unsure if it is correct that count in that same method starts at 1.

In java [RuntimeException](https://docs.oracle.com/javase/8/docs/api/java/lang/RuntimeException.html) inherits from [Exception](https://docs.oracle.com/javase/8/docs/api/java/lang/Exception.html) which in turn is an implementation of [Throwable](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html). This should allow us to aggregate all RuntimeExceptions (uncheck exceptions) that could be thrown from within the lambda. As for the count start value, if I'm [not mistaken](http://docstore.mik.ua/orelly/java-ent/jnut/ch02_05.htm#javanut3-ch-2-tab-5) reading from var++ means read the existing value then increment. Whereas ++var means increment, then read.

> I'm not sure Part benefits from being an inner class in the long run. (Again I'm unsure of the intended longevity of this Java API.)

I took your advice and moved it into its own class.

> I suspect validateThereAreNoMissingParts is going to be hard to use as written. MantaClientException wouldn't be what I would initially expect for missing parts.

You are right. This needs its own exception type. I'm glad you caught that. Also, I'm a bit on the fence about the utility of this method when we have the server side implementation. @jordanhendricks, do you have any thoughts?

> In the spirit of Joyent conventions, should MultipartMetadata be serialized as JSON?

That implementation will be thrown away once we have the server side implementation and it is a bit of work for me to add the serialization logic for this with the aggressive deadline. Do you think it is worth it?

> In waitForCompletion, what happens if we exceed timesToPoll?

Great catch. We just stop waiting and there is no recourse to the caller. I'm thinking that we will add a lambda to allow the caller to define their own behavior.

> I know this is a tide-over POC, but if I mrm one of the parts mid upload, the commit still succeeded.

Well, that mean's that the job didn't fail. I need to investigate what happens when you mget A B C and B doesn't exist. This might be out of scope.

> Tried to upload 9000 files, but the commit failed with {"code":"RequestEntityTooLargeError","message":"Request body size exceeds 100000"} I don't think the string builder can handle near 10k (maybe that's okay initially?)

This isn't a problem with StringBuilder. Rather it is a problem with the Manta job definition being too large. I don't think there is anything that I can do about this.

> I'm worried that the current semantics of committing ("cat everything in the upload dir together") are significantly different from what we want the final API to be ("commit these etags").

I'm working on this problem right now. Stay tuned.

cburroughs:
> My intention was to throw away the jobs based implementation entirely once we add support on the server. This would allow for little to no code changes on the part of the consumer of the Java SDK. Whereas if we did the interface to multiple implementation method, the consumer would have to change their code.

I'm confused here.  If I'm using `MantaMultipartManager` and it goes away, don't I need to change my code to use the New Thing?  If the New Thing is like `MantaMultipartManager`, then isn't that like having an interface?

> This is probably a deficiency in the SDK itself. Also, that error isn't typical for an authentication failed error. Can you do a PR on this branch adding an integration tests that causes that message to be produced?

I believe this is unrelated to the MPU work and opened #123

> They don't need to be serializable once we have support at the server. However, I do use Java serialization to store state on Manta for the jobs-based implementation. I don't think it hurts anything. Also, there is a cult among the larger Java religion that believes that all classes should be marked as Serializable unless there are truly intended not to be. That is why classes like Exception implement Serializable - I mean who wants to serialize exceptions?!? I have no stake in this game.

Ah, well I guess I am a member of the rival Java Serialization is Terrible and should be a last resort cult.  Whatever, consistency makes sense with the rest of `java-manta`

>> If we are already depending on a library that can generate them, a time based UUID might help debug some odd corner case where mtime is insufficient down the road.

>I don't quite understand what you mean. Could you explain this further?

With a time based UUID we would have a clue to when the MPU was created, and not just the `mtime` (which can change, right?) of the dir.  But java inexplicably can't generate time based UUIDs with the standard library, and it's probably not worth adding new dependencies just for that.

Regarding the error handling.  I get the aggregation of `java.lang.Exception` and I think `List<Exception> exceptions = new ArrayList<>();` it would be fine.  But having `List<Throwable>` makes it look like we are *also* trying to catch `java.lang.Error`, which is almost always bad.

>> In the spirit of Joyent conventions, should MultipartMetadata be serialized as JSON?

> That implementation will be thrown away once we have the server side implementation and it is a bit of work for me to add the serialization logic for this with the aggressive deadline. Do you think it is worth it?

Given the ephemeral nature I'd defer to you on how heavily you think this will be used and how likely that humans will be debugging.

> This isn't a problem with StringBuilder. Rather it is a problem with the Manta job definition being too large. I don't think there is anything that I can do about this.

Would it make sense to lower `MAX_PARTS` to something that is more likely to succeed? I don't see a point to allowing uploads we know will fail.

 * I'm unsure about `MantaMultiPartManager.complete` taking a Stream. (Disclosure: Most of my java experience was 6/7.)  Isn't a `List` still the idiomatic way to keep track of a list of things? (And it looks like that's what you do in the readme example).
 * Now with aborting I get `com.joyent.manta.exception.MantaMultipartException: The multipart upload specified hasn't been committed/completed yet`.  It should be possible to abort an upload, and not just abort the under-the-hood-job, right?

Overall I think this is a closer match the the server-side API as I understand it.  I have some questions outstanding to make sure my understanding also matches reality.  If that goes well I can do one more pass early next week looking for line-by-line nits.

dekobon:
> I'm confused here. If I'm using MantaMultipartManager and it goes away, don't I need to change my code to use the New Thing? If the New Thing is like MantaMultipartManager, then isn't that like having an interface?

Yes. The point was to use it like an interface (in the general sense). Rather than comparing it to Java's class type interface, you could compare it to a [JSR](https://www.jcp.org/en/jsr/overview). JSRs outline the full implementation API for a given set of functionality in Java. Library developers code their libraries such that they conform exactly to the JSR class names and packages. Then consumers of the library can add in a jar file to get the functionality of the JSR. If they don't like the library, they can swap it out for a different implementation of the JSR without changing their code. The canonical example of a JSR that is swappable is the [JSR 315 - Servlet 3.0 Specification](https://jcp.org/en/jsr/detail?id=315). Tomcat, Jetty and WebLogic all have their own implementations of the classes defined in the JSR. However, from an application developer's perspective those classes have a constant API and theoretically (and often in practice) you can swap out the JSR implementation and your application will still function.

In summary, we want to give the Samsung team a nearly final Java API that they will not need to change (much). Right now they don't care about the latencies or the scalability. They just need the multipart API to be defined firmly so that they can make progress on their Manta/S3 abstraction layer. If switched up the class names, then it would defeat the purpose.

That said, I've written a lot for how much of a weak feeling I have towards going to the interface / multiple implementation route. My main thrust is that I want to minimize all changes. Even the difference between ```new MantaJobMultipartManger()``` and ```new MantaNativeMultipartManager()``` because those all cause additional speed bumps in a large company.

> I believe this is unrelated to the MPU work and opened #123

Thank you for doing this.

> Ah, well I guess I am a member of the rival Java Serialization is Terrible and should be a last resort cult. 

Honestly, I'm a member of the same camp, but maybe my torch isn't lit... I was just lazy. I switched it over to use json.

> With a time based UUID we would have a clue to when the MPU was created, and not just the mtime (which can change, right?) of the dir. But java inexplicably can't generate time based UUIDs with the standard library, and it's probably not worth adding new dependencies just for that.

com.fasterxml (the json jackson folks) have a nice UUID library that fits in the driver well. I switched to using it for generating time-based UUIDs.

> Regarding the error handling. I get the aggregation of java.lang.Exception and I think List<Exception> exceptions = new ArrayList<>(); it would be fine. But having List<Throwable> makes it look like we are also trying to catch java.lang.Error, which is almost always bad.

I switched to using Exception instead of Throwable.

> Would it make sense to lower MAX_PARTS to something that is more likely to succeed? I don't see a point to allowing uploads we know will fail.

I did some back of the napkin math about average object filename sizes and came up with a value for 1,000 for MAX_PARTS. We will up it when there is native multipart support.

> I'm unsure about MantaMultiPartManager.complete taking a Stream. (Disclosure: Most of my java experience was 6/7.) Isn't a List still the idiomatic way to keep track of a list of things? (And it looks like that's what you do in the readme example).

Collection/List/Set/Map are all in-memory data structures that don't allow you to stream data. There has been a movement for sometime to get everyone to code their APIs in the minimal type needed to process the inputs. In Java 6/7 this looked like Iterator or Iterable in the APIs. With Java 8 we have the additional type of Stream. I do agree that having just a Stream there is a bit of a weird API, so I added support for Iterable and you can now just drop a Collection/List/Set in the API.

> Now with aborting I get com.joyent.manta.exception.MantaMultipartException: The multipart upload specified hasn't been committed/completed yet. It should be possible to abort an upload, and not just abort the under-the-hood-job, right?

You mentioned this and I totally misunderstood my own thinking about it. I think I was tired. This *should* be fixed now.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/MantaMultipartManager.java
```diff
@@ -0,0 +1,1106 @@
...
+        }
+
+        throw aggregateException;
+    }
+
+    /**
+     * Initializes a new multipart upload for an object.
+     *
+     * @param path remote path of Manta object to be uploaded
+     * @return unique id for the multipart upload
```
cburroughs: nit: Makes it sound like the return type is a uuid instead of `MantaMultipartUpload`?

java-manta-client/src/main/java/com/joyent/manta/client/multipart/MantaMultipartManager.java
```diff
@@ -0,0 +1,1106 @@
...
+            MantaMultipartException e = new MantaMultipartException(msg);
+            e.setContextValue("upload_id", id.toString());
+            e.setContextValue("job_id", id.toString());
+
+            throw e;
+        }
+
+        MantaJobBuilder.Run run = mantaClient.jobBuilder().lookupJob(job);
+
+        if (LOG.isDebugEnabled()) {
```
cburroughs: FWIW I don't think these extra guards are needed if they hinder readability.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -278,17 +286,27 @@ public void deleteRecursive(final String path) throws IOException {
...
-                LOG.warn("First attempt to delete directory failed, retrying");
-                // Re-attempt to delete the directory
-                this.deleteRecursive(path);
-            } catch (InterruptedException ie) {
-                // We don't need to do anything, just exit
+        } catch (MantaClientHttpResponseException e) {
+            // Directory wasn't empty
+            if (e.getServerCode().equals(DIRECTORY_NOT_EMPTY_ERROR)) {
+                try {
+                    final int waitTime = 400;
```
dekobon: Sometimes we hit a race condition when recursively deleting. Many times, there is another recursive delete operation also running. Giving it a moment to wait for the other process to stop increases the likelihood that the retry succeeds. When you reason about it theoretically, this makes no sense. When you look at how all of the SDKs are implemented and how most users implement them, they tend to fall in this pattern and this becomes a pragmatic choice.
Honestly, this design decision needs review and revisiting, so I've created [this issue](https://github.com/joyent/java-manta/issues/125).

java-manta-client/src/main/java/com/joyent/manta/client/multipart/MantaMultipartManager.java
```diff
@@ -0,0 +1,1106 @@
...
+    static final String MULTIPART_DIRECTORY =
+            "stor/.multipart-6439b444-9041-11e6-9be2-9f622f483d01";
+
+    /**
+     * Metadata file containing information about final multipart file.
+     */
+    static final String METADATA_FILE = "metadata.json";
+
+    /**
+     * Default number of seconds to poll Manta to see if a job is complete.
```
dekobon: Good nit. I removed "Default".

java-manta-client/src/main/java/com/joyent/manta/client/multipart/MantaMultipartManager.java
dekobon: Another good nit. I fixed this.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/MantaMultipartManager.java
dekobon: This guard is needed because we are doing a networking call to get information for the debug message. Without it, it would be executed with every invocation.

dekobon:
@cburroughs Thank you so much for the code reviews. I've been looking forward to getting a code review at Joyent for the last year. This has been very educational and useful.

Issue #137, opened 2017-01-01
Add get method with explicit range parameters


dekobon:
Currently, if you want to pull a specific byte range from Manta you have to craft a custom MantaHttpHeaders object containing the [Range HTTP header](https://tools.ietf.org/html/rfc7233) or you have to get an instance of MantaSeekableByteChannel.

This feature request is to create a new method on MantaClient that allows you specify a byte range to get and it creates the HTTP Range header for you.

cburroughs:
There are currently a plethora of get methods:

 * `MantaObjectResponse get(final String rawPath)`
 * `MantaObjectInputStream getAsInputStream(final String rawPath, final MantaHttpHeaders requestHeaders)`
 * `MantaObjectInputStream getAsInputStream(final String path)`
 * `String getAsString(final String path)`
 * `String getAsString(final String path, final String charsetName)`
 * `String getAsString(final String path, final Charset charset)`
 * `Path getToTempPath(final String path)`
 * `File getToTempFile(final String path)`
 * `MantaSeekableByteChannel getSeekableByteChannel(final String rawPath, final long position)`
 * `MantaSeekableByteChannel getSeekableByteChannel(final String rawPath)`

Did you have a thought as to how you wanted the api to look?  Should there be a companion `getRangeAsFoo` for each existing method?

dekobon:
I envision only adding this method:

```java
MantaObjectInputStream getAsInputStream(String rawPath, MantaHttpHeaders requestHeaders, long startPosition, long endPosition)
```

Many of the other methods are convenience methods that allow consumers of the API use a lot less code/boilerplate to interact with it.

PR #140, opened 2017-01-24
provide a get method with explicit range parameters

cburroughs:
Users who want multiple ranges will still need to manipulate their own
headers.

NOTE: This also enables all of the `com.joyent.manta.http` tests which
were lying dormant.

ref #137


dekobon:
LGTM

PR #141, opened 2017-01-24
bind checkstyle to compile instead of validate phase

cburroughs:
The `validate` phase occurs *before* compilation, which means that if
there was a syntax error checkstyle would die trying to parse the file
instead of javac just telling you what the syntax error is. Binding it
to `compile` (instead of the perhaps more descriptive `verify`)
preserves the aggressive & early nature of the check.


dekobon:
Great change.

PR #148, opened 2017-01-28
Range request api

dekobon:
This PR adds support for range request with the AES/CTR cipher/mode.


java-manta-client/src/main/java/com/joyent/manta/http/EncryptionHttpHelper.java
```diff
@@ -278,9 +294,116 @@ public MantaObjectInputStream httpRequestAsInputStream(final HttpUriRequest requ
...
+     * @param request source request that hasn't been made yet
+     * @param requestHeaders headers passed to the request
+     * @return a {@link Long} array containing two elements: skip bytes, plaintext length
+     * @throws IOException thrown when we fail making an additional HEAD request
+     */
+    @SuppressWarnings("MagicNumber")
+    private Long[] calculateSkipBytesAndPlaintextLength(final HttpUriRequest request,
+                                                        final MantaHttpHeaders requestHeaders)
+            throws IOException {
+        // TODO: Unhardcode cipher
```
dekobon: We need to discuss the design how we decrypt files.

geek:
LGTM

java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
```diff
@@ -368,6 +506,26 @@ public long skip(final long numberOfBytesToSkip) throws IOException {
             throw e;
         }
 
+        skipInitialBytes();
+
+        /* I would really love to actually use the CipherInputStream.skip() method
+         * when we aren't using authenticated encryption or we are using a AEAD cipher,
+         * but the CipherInputStream method is horribly broken in some cases
+         * and doesn't return back the actual number of bytes skipped. */
```
cburroughs: This is useful information!  Please reword the tone a bit into a comment we can keep.  How else is `skip` broken?

java-manta-client/src/main/java/com/joyent/manta/http/EncryptionHttpHelper.java
```diff
@@ -866,4 +989,34 @@ private Cipher buildMetadataDecryptCipher(final SupportedCipherDetails cipherDet
         }
         return metadataCipher;
     }
+
+    /**
+     * Converts a nullable {@link Long} array into a long primitive array.
```
cburroughs: Is this saying that `ranges` can be `null`? Or that some of the values can be `null`?

java-manta-client/src/main/java/com/joyent/manta/http/EncryptionHttpHelper.java
```diff
@@ -866,4 +989,34 @@ private Cipher buildMetadataDecryptCipher(final SupportedCipherDetails cipherDet
...
+     * Unlimited positions are represented as the hard file size limits of the
+     * specified cipher.
+     *
+     * @param ranges array containing two elements
+     * @param cipherDetails cipher being used to compute range
+     * @return updated range coordinates based on cipher/cipher mode configuration
+     */
+    private static long[] byteRangeAsNullSafe(final Long[] ranges,
```
cburroughs: Can this be `@VisibleForTesting` (even if we are not using guava or whatever has the actual formal annotation)?

java-manta-client/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
```diff
@@ -284,15 +274,16 @@ public long size() throws IOException {
...
+        final MantaObjectInputStream stream = connectOrGetResponse();
+        final long contentLength = stream.getContentLength();
 
         if (contentLength == UNKNOWN_CONTENT_LENGTH) {
             MantaClientException e = new MantaClientException(
                     "Can't get SeekableByteChannel for objects of unknown size");
             HttpUriRequest request = requestRef.get();
             if (request != null) {
+                @SuppressWarnings("unchecked")
+                HttpResponse response = (HttpResponse)stream.getHttpResponse();
```
dekobon: This is returned as an untyped object because when building the JAR, we shade Apache HTTP Client. This will change the object signature to com.joyent.org.apache.http.*

In order to get this value to be passed between packages, it needs to be public, but we don't want to expose the underlying API for the library because it may change out from under people's feet in the future. I've seen this pattern before. Untyped objects are occasionally available publicly for debugging if an implementor wants it, but there are no guarantees as to what object will actually be there.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
```diff
@@ -63,7 +59,15 @@ public long ciphertextSize(final long plaintextSize) {
             calculatedContentLength = (plaintextSize - padding) + blockBytes;
         }
 
-        // Append a block of 0's at the end of stream
+        byte[] iv = this.getCipher().getIV();
+        // CBC requires an IV an extra block for chaining to work
+        if ((iv == null || iv.length == 0) && plaintextSize >= blockBytes) {
+            final double blocks = Math.floor(plaintextSize / ((long) blockBytes));
+            calculatedContentLength = (((long)blocks + 1) * blockBytes);
+        }
```
dekobon: @geek Can you handle this?

java-manta-client/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
```diff
@@ -117,20 +115,20 @@ public MantaSeekableByteChannel(final String path,
      * channel from within this class. This is used when position() is called.
      *
      * @param requestRef reference to existing HTTP request
-     * @param responseRef reference to existing HTTP response
+     * @param responseStream reference to existing HTTP response
```
dekobon: Yes, that is correct.

java-manta-client/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
```diff
@@ -178,11 +174,14 @@ public int read() throws IOException {
...
-        final HttpResponse response = connectOrGetResponse();
-        final HttpEntity entity = response.getEntity();
-        final InputStream is = entity.getContent();
-        position++;
-        return is.read();
+        final MantaObjectInputStream stream = connectOrGetResponse();
+        final int read = stream.read();
+
+        if (read > -1) {
+            position++;
```
dekobon: Great catch. I got atomicity and memory visibly confused. For some reason, I misremembered volatile as AtomicX for primitives. This is certainly not true. I've updated to use a AtomicLong.

java-manta-client/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
```diff
@@ -347,33 +338,36 @@ public synchronized void close() throws IOException {
...
+        if (responseStream.get() != null) {
+            return responseStream.get();
         }
 
         final HttpUriRequest request = connectionFactory.get(path);
+        final MantaHttpHeaders headers = new MantaHttpHeaders();
 
         // Set byte range requested via HTTP range header
-        request.setHeader(HttpHeaders.RANGE, String.format("bytes=%d-", position));
+        headers.setRange(String.format("bytes=%d-", position));
```
dekobon: That is the intention. I believe that there should be a hanging HTTP connection open. It will progressively read the bytes over that channel. If the consumer wants to end early, the close() call will break the connection.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
```diff
@@ -51,7 +52,7 @@
     /**
      * Default buffer size to use when reading chunks of data from streams.
      */
-    private static final int DEFAULT_BUFFER_SIZE = 8192;
+    private static final int DEFAULT_BUFFER_SIZE = 512;
```
dekobon: I did some benchmarking and the CipherInputStream doesn't benefit from a greater buffer size.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCtrCipherDetails.java
```diff
@@ -59,42 +61,82 @@ public boolean plaintextSizeCalculationIsAnEstimate() {
 
     @Override
     public long[] translateByteRange(final long startInclusive, final long endInclusive) {
```
dekobon: Let's do that once we integrate. My thoughts as well. I've created #159 to capture that work.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
```diff
@@ -76,7 +77,7 @@
     /**
      * The cipher stream that wraps the backing stream allowing for streaming decryption.
      */
-    private final CipherInputStream cipherInputStream;
+    private final InputStream cipherInputStream;
```
dekobon: Fixed.

java-manta-client/src/main/java/com/joyent/manta/http/EncryptionHttpHelper.java
dekobon: I changed this to package scope. @geek can you add some additional testing to verify this function's behavior?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
```diff
@@ -94,22 +95,114 @@
...
 
     /**
+     * Flag indicating if we perform authentication on the ciphertext.
+     */
+    private final boolean authenticateCiphertext;
+
+    /**
+     * Starting position in plaintext.
+     */
+    private final Long startPosition;
```
dekobon: Added a comment.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
```diff
@@ -94,22 +95,114 @@
...
+        initializeHmac();
+    }
+
+    /**
+     * Verifies that the plaintext length supplied is less than or equal to the
+     * size of the original plaintext object. If it is sized improperly, we
+     * attempt to convert the plaintext length to the size of the plaintext object.
+     * If we have insufficient information, the plaintext value returned may
+     * be inaccurate, so it is ultimately the responsibility of the caller to
+     * make sure that this value is accurate.
```
dekobon: I imagined a number of cases where the consumer of the SDK would specify a byte range that exceeds the size of the end of the file. I've seen this handled before by just defaulting to the end of the file. It seems rather draconian to insist that the value is correct if we can cope with it. In fact, HTTP range requests operate in this model as well.

The autocorrect is exercised in the following unit tests:
* `MantaClientRangedIT.canGetWithEndRangeBeyondObjectSize`
* `MantaClientRangedIT.canGetWithUnboundedEndRange`

The first one (`MantaClientRangedIT.canGetWithEndRangeBeyondObjectSize`) is an obvious use of it. The second one is interesting. In order to do unbounded ranges, I just set a range to the maximum ciphertext size.

I went and removed the comment about the value being inaccurate because in the `HttpHelper.attemptToFindPlaintextSize()` method we explicitly fail if we are unable to calculate an accurate value.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
dekobon: I'll leave out the folksy tone... Fixed.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
geek: https://github.com/joyent/java-manta/pull/162

PR #162, opened 2017-01-30
Testing around CBC content-length calc

geek:


java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
```diff
@@ -104,4 +78,43 @@ public long updateCipherToPosition(final Cipher cipher, final long position) {
...
     }
+
+    /**
+     *
+     * @param plaintextSize size in bytes of unencrypted body
+     * @param blockBytes block size in bytes
+     * @param tagOrHmacBytes number of bytes needed for the hmac or authentication tag
+     * @return calculated byte size of encrypted body with the hmac/authentication tag appended
+     */
+    public static long calculateContentLength(final long plaintextSize, final int blockBytes, final int tagOrHmacBytes, final boolean hasIV) {
```
dekobon: You can change the scope of the method to the default scope by removing `public`. I think we should because this method is only used within this class. By changing to the default scope access is limited to only within the package. This allows your tests to still run and to limit scope without having to mark it as `private`.

PR #220, opened 2017-03-15
support multiple singing key types 

cburroughs:
Using the new support in java-http-signature 4.0.0, support DSA and
ECDSA keys.  A user can provide a key and the library will figure out
it's type and use sensible defaults.

As part of this I had to tease apart how `ThreadLocalSigner` objects
are managed by centralizing their creation in `MantaClient`.  It was
too easy to end up with multiple ones with different configuration
which only accidental worked when everything was RSA.  While several
class changed, the public api of `MantaClient` did not.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
         <maven-error-prone-core.version>2.0.15</maven-error-prone-core.version>
 
         <!-- Dependency versions -->
-        <dependency.http-client-signature.version>3.0.2</dependency.http-client-signature.version>
+        <dependency.http-client-signature.version>4.0.1-SNAPSHOT</dependency.http-client-signature.version>
         <dependency.apache-http-client.version>4.5.3</dependency.apache-http-client.version>
         <dependency.bouncycastle.version>1.55</dependency.bouncycastle.version>
         <dependency.fasterxml-uuid>3.1.3</dependency.fasterxml-uuid>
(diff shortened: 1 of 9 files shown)


cburroughs:
NOTE: That this depend on `SNAPSHOT`, so I'm not sure when you want to land this relative to the release.

cburroughs:
I see one place in UriSigner that still has a hard coded RSA reference I need to figure out.

cburroughs:
> I see one place in UriSigner that still has a hard coded RSA reference I need to figure out.

Okay I think I see the (small) change that has to happen, but it will require a change to http-sig to expose a string like "rsa-sha256".  I think I just never got to "signer" when I left the integration tests running for a while.  I'll should be able to get those diffs up this afternoon.

pom.xml
```diff
@@ -252,6 +252,10 @@
                          only supports >=6 -->
                     <source>${java.min.version}</source>
                     <target>${java.min.version}</target>
+                    <compilerArgs>
```
dekobon: We might as well set: `-Xlint:all`

cburroughs:
With the latest diff (and latests http-sig) I could run all ` -Dit.test=MantaClient*IT`, so I hope there are no more surprises.

cburroughs:
(I think this is ready to land, but the hope was to get a release out of java-http-signature first)

dekobon:
Yes, let's cut a version of the latest HTTP signatures and then I will run an integration test of this branch. If all goes well, then I will push it out as a snapshot.

PR #224, opened 2017-04-03
detect last MPU part to maintain min size

cburroughs:
Previously the buffer and HAMC bytes were always unconditionally
written as a final part.  This meant that if the last user generated
part was less than the minimum part size, two parts below the minimum
would be uploaded.

To avoid this, keep track of the bytes for uploaded for each part. If
less than the minimum bytes have been uploaded then this must be the
last part (otherwise there would be an error on completion).  If that
is the case append the remaining cipher bytes to the last part, and do
*not* create an additional part during completion.

ref #218

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionState.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionState.java
@@ -8,7 +8,16 @@
 package com.joyent.manta.client.multipart;
 
 import com.joyent.manta.client.crypto.EncryptionContext;
+import com.joyent.manta.exception.MantaMultipartException;
+import com.joyent.manta.util.HmacOutputStream;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang3.Validate;
+import org.bouncycastle.crypto.macs.HMac;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Objects;
 import java.util.concurrent.locks.ReentrantLock;
@@ -21,6 +30,12 @@ import java.util.concurrent.locks.ReentrantLock;
  * @since 3.0.0
  */
 public class EncryptionState {
+
+    /**
+     * Logger instance.
+     */
+    private static final transient Logger LOGGER = LoggerFactory.getLogger(EncryptionState.class);
+
     /**
      * Encryption cipher state object.
      */
@@ -45,6 +60,12 @@ public class EncryptionState {
      * The encrypting stream.
      */
     private transient OutputStream cipherStream = null;
+
+    /**
+     * Indicates if the multipart stream buffer has been flushed and
+     * final cipher auth bytes written.
+     */
+    private boolean lastPartAuthWritten = false;
 
     /**
      * Zero argument constructor used for serialization.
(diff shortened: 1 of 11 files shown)


cburroughs:
I've run all of the integration tests.  The serialization tests
pass (?) without any changes.  I'm not sure what the consequences of the
field addition are there.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionState.java
```diff
@@ -47,6 +62,12 @@
     private transient OutputStream cipherStream = null;
 
     /**
+     * Indicates if the multipart stream buffer has been flushed and
+     * final cipher auth bytes written.
+     */
+    private boolean lastPartAuthWritten = false;
```
dekobon: I suggest that you incorporate this field into the serialized state:
```
diff --git a/java-manta-client-kryo-serialization/src/main/java/com/joyent/manta/serialization/EncryptionStateSerializer.java b/java-manta-client-kryo-serialization/src/main/java/com/joyent/manta/serialization/EncryptionStateSerializer.java
index 8d5ad7e..3e1be0d 100644
--- a/java-manta-client-kryo-serialization/src/main/java/com/joyent/manta/serialization/EncryptionStateSerializer.java
+++ b/java-manta-client-kryo-serialization/src/main/java/com/joyent/manta/serialization/EncryptionStateSerializer.java
@@ -45,6 +45,7 @@ public class EncryptionStateSerializer extends AbstractManualSerializer<Encrypti
     private Field lastPartNumberField = captureField("lastPartNumber");
     private Field multipartStreamField = captureField("multipartStream");
     private Field cipherStreamField = captureField("cipherStream");
+    private Field lastPartAuthWrittenField = captureField("lastPartAuthWritten");
 
     /**
      * Secret key to inject into encryption context.
@@ -119,6 +120,9 @@ public class EncryptionStateSerializer extends AbstractManualSerializer<Encrypti
         final int lastPartNumber = (int)readField(lastPartNumberField, object);
         output.writeInt(lastPartNumber, true);
 
+        final boolean lastPartAuthWritten = (boolean)readField(lastPartAuthWrittenField, object);
+        output.writeBoolean(lastPartAuthWritten);
+
         MultipartOutputStream multipartStream = (MultipartOutputStream)readField(multipartStreamField, object);
         ByteArrayOutputStream multipartStreamBuf;
 
@@ -155,10 +159,13 @@ public class EncryptionStateSerializer extends AbstractManualSerializer<Encrypti
         final int lastPartNumber = input.readVarInt(true);
         encryptionContext.setKey(secretKey);
 
+        final boolean lastPartAuthWritten = input.readBoolean();
+
         final EncryptionState encryptionState = newInstance();
 
         writeField(encryptionContextField, encryptionState, encryptionContext);
         writeField(lastPartNumberField, encryptionState, lastPartNumber);
+        writeField(lastPartAuthWrittenField, encryptionState, lastPartAuthWritten);
 
         final int blockSize = encryptionContext.getCipherDetails().getBlockSizeInBytes();
         Object bufferObject = kryo.readClassAndObject(input);
```

cburroughs:
I'm still really fuzzy on how Kryo works.  In `EncryptionStateSerializerTest` I set `lastPartAuthWritten` to a non-default value, serialize it, deserialize it, and get the same value back.  What would the explicit changes in `EncryptionStateSerializer` do differently?  How can I test that whatever is different is working?

dekobon:
I wonder if there was a bug in the diff that I sent you. Could you check in your failed test? I'll take a look at it. Kyro is highly finicky and sometimes is flakey when certain data types are located side by side.

cburroughs:
I don't have a failed test, which is part of what leaves me so confused.  `EncryptionStateSerializerTest` *passes* with no changes to  `EncryptionStateSerializer`. ¯\_(ツ)_/¯

dekobon:
The test is rather light in that it uses the equals method on `EncryptionState` to see if the two objects are equivalent. Did you add `lastPartAuthWritten` to the equals method?

cburroughs:
>  Did you add lastPartAuthWritten to the equals method?

Yep. https://github.com/joyent/java-manta/pull/224/files#diff-774387bf947c7b5f619097e41fba5b58R175

cburroughs:
From a side conversation: The test worked because the test wasn't using `EncryptionStateSerializer` at all but the default Kryo one.  I fixed the test, saw it fail, and incorporated the above suggested changes (and the test then passed).

It's possible that `EncryptionStateSerializer` isn't needed at all, but since it's the one that has been tested and evaluated the consensus was to stick with it/

Issue #223, opened 2017-03-20
Add configurable buffer size for streaming copy operations


dekobon:
Initially I found that smaller buffers provided better performance with cryptographic operations in pure Java. However, upon additional tweaking, I've found that there are performance benefits from tweaking buffer sizes when using libnss.

This issue is to create a configuration parameter that allows the following values to be configurable:
[MantaEncryptedObjectInputStream](https://github.com/joyent/java-manta/blob/3.0/java-manta-client/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java#L68)
[EncryptingEntity](https://github.com/joyent/java-manta/blob/3.0/java-manta-client/src/main/java/com/joyent/manta/client/crypto/EncryptingEntity.java#L195)
[EncryptingPartEntity](https://github.com/joyent/java-manta/blob/3.0/java-manta-client/src/main/java/com/joyent/manta/client/crypto/EncryptingPartEntity.java#L47)

cburroughs:
Do you anticipate *3* different parameters?  Or just a single encrypted buffer param?  If 3, why would a user choose different values?

dekobon:
I wanted you to give it some thought. I *think* one setting would be sufficient because it is sizing the byte array used as a buffer for copying streams. 

According to @arekinath a higher value would yield better performance when interacting with AES-NI. However, since the streams are chained to operations going over a network socket, I'm suspicious about it actually getting better performance with higher values.

cburroughs:
I've extracted `EncryptingEntityBenchmark` into a JMH benchmark and made the `EncryptingEntity` bufferSize configurable.  Looking specifically at encrypting 1 MiB:

```
BC
Benchmark                                (bufferSize)  (encryptionAlgorithm)   Mode  Cnt   Score   Error  Units
EncryptingEntityBenchmark.encryptOneMiB            32   AES128/CTR/NoPadding  thrpt   60  32.694 ± 0.223  ops/s
EncryptingEntityBenchmark.encryptOneMiB            64   AES128/CTR/NoPadding  thrpt   60  33.880 ± 0.303  ops/s
EncryptingEntityBenchmark.encryptOneMiB           128   AES128/CTR/NoPadding  thrpt   60  34.971 ± 0.240  ops/s
EncryptingEntityBenchmark.encryptOneMiB           256   AES128/CTR/NoPadding  thrpt   60  35.856 ± 0.067  ops/s
[...]

PR #225, opened 2017-04-04
convert EncryptingEntity benchmark to use JMH

cburroughs:
JMH helps make better Java benchmarks that avoid common pitfalls.

NOTE: This adds scaffolding for buffer size tests, but does not make
it configurable since measurements showed it was not worthwhile to do
so.

ref #223

### DIFF
--- a/java-manta-benchmark/pom.xml
+++ b/java-manta-benchmark/pom.xml
@@ -19,6 +19,8 @@
     <packaging>jar</packaging>
 
     <properties>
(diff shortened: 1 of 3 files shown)


cburroughs:
I think anything that's a "micro" benchmark (consumes CPU and memory, but minimal interaction with the outside world) is a very good fit for JMH and I have been happy with it.  I don't think it would help -- and would probably get in the way -- with "integration" benchmarks that interacted with disk, network, or Manta.

PR #228, opened 2017-04-06
reduce test failure noise

cburroughs:
Since async commits are not part of MPU, these theses do not need to
be so asynchronously clever.  Errors in the threadpool or abort code
lead to inconsistent failures.

Also add some missing toString methods to aid in test debugging.


java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
```diff
@@ -431,6 +434,7 @@ public void cantUploadSmallMultipartString() throws IOException {
                                         });
             assert exception.getCause() instanceof IllegalStateException;
         }
+        Assert.assertEquals(uploadedParts.size(), 1, "only first small upload should succeed");
```
cburroughs: I added this for an error that I got once and have not been able to reproduce.  I'm hopeful that this assertion will make me feel less crazy.

dekobon: That makes sense to me.

Issue #230, opened 2017-04-17
some MantaClient.put paths skip url-encoding


cburroughs:
ex, for a `File`:

```
 $ java -jar java-manta-cli/target/java-manta-cli-3.0.1-SNAPSHOT-jar-with-dependencies.jar put-file /tmp/hi '/cburroughs/stor/tmp/!@#$%^&*()'
14:57:28.839 [main] DEBUG com.joyent.manta.client.crypto.ExternalSecurityProviderLoader - Bouncy Castle provider was not loaded, adding to providers
14:57:28.843 [main] DEBUG com.joyent.manta.client.MantaClient - Preferred Security Provider: BC version 1.56
14:57:29.154 [main] DEBUG com.joyent.manta.http.StandardHttpHelper - PUT    /cburroughs/stor/tmp/!@#$%^&*()
14:57:29.161 [main] DEBUG com.joyent.manta.org.apache.http.impl.conn.PoolingHttpClientConnectionManager - Connection manager is shutting down
14:57:29.162 [main] DEBUG com.joyent.manta.org.apache.http.impl.conn.PoolingHttpClientConnectionManager - Connection manager shut down
Malformed escape pair at index 57: https://us-east.manta.joyent.com/cburroughs/stor/tmp/!@#$%^&*()
java.lang.IllegalArgumentException: Malformed escape pair at index 57: https://us-east.manta.joyent.com/cburroughs/stor/tmp/!@#$%^&*()
	at java.net.URI.create(URI.java:852)
	at com.joyent.manta.org.apache.http.client.methods.HttpPut.<init>(HttpPut.java:65)
	at com.joyent.manta.http.MantaConnectionFactory.put(MantaConnectionFactory.java:602)
	at com.joyent.manta.http.StandardHttpHelper.httpPut(StandardHttpHelper.java:177)
	at com.joyent.manta.client.MantaClient.put(MantaClient.java:1216)
	at com.joyent.manta.client.MantaClient.put(MantaClient.java:1146)
	at com.joyent.manta.cli.MantaCLI.putFile(MantaCLI.java:282)
	at com.joyent.manta.cli.MantaCLI.main(MantaCLI.java:104)
Caused by: java.net.URISyntaxException: Malformed escape pair at index 57: https://us-east.manta.joyent.com/cburroughs/stor/tmp/!@#$%^&*()
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.scanEscape(URI.java:2978)
[...]

Issue #229, opened 2017-04-16
Problems with handling spaces in file/directory names


phillipross:
I have files and directory stored in manta that actually have spaces.  It seems node-manta tools allow this (I populated the directories with `muntar`).  The node-manta tools `mls` and `mfind` properly list the files with spaces too:
```
$ mls ~~/stor/ica/images_2016_1022/abh  | grep " " | head -2
Adelaide images/
April 2014/
$ mfind ~~/stor/ica/images_2016_1022/abh  | grep " " | head -2
/philross/stor/ica/images_2016_1022/abh/Adelaide images
/philross/stor/ica/images_2016_1022/abh/April 2014
```
Unfortunately, the java-manta client seems to return these paths with spaces replaced by plus symbols.  Trying to request one of these paths then fails as the path doesn't actually exist with plus sybols... it exists with spaces.

The solution isn't as simple as `URLEncode.encode(path, 'UTF8')` either since that will replace the path separators with %2F.

phillipross:
It appears [MantaUtils.formatPath](https://github.com/joyent/java-manta/blob/master/java-manta-client/src/main/java/com/joyent/manta/util/MantaUtils.java) method is using `URLEncode.encode` which will change the spaces to plus symbols and manta will store the paths with these characters as plus symbols.  I think they should be encoded to `%20` instead of the plus symbol.

The following shows a series of tests I did with curl:
```
$ manta "/$MANTA_USER/stor/java-manta-testing" -XPUT -H"content-type: application/json; type=directory"
$ manta "/$MANTA_USER/stor/java-manta-testing"
$ manta "/$MANTA_USER/stor/java-manta-testing/a b c" -XPUT -H"content-type: application/json; type=directory"
{"BadHTTPRequest": "Request was malformed"}
$ manta "/$MANTA_USER/stor/java-manta-testing/a+b%20c" -XPUT -H"content-type: application/json; type=directory"
$ manta "/$MANTA_USER/stor/java-manta-testing"
{"name":"a+b c","type":"directory","mtime":"2017-04-16T04:44:05.852Z"}
```
In the output, you can see that creating directories with spaces in the path makes manta unhappy.  But plus symbols are indeed stored as plus symbols, while %20 chars are stored as spaces.

cburroughs:
Thank you for the detailed report.  In `master` spaces are now encoded as spaces (and match the node behavior).

PR #234, opened 2017-04-21
consistently urlencode paths correctly

cburroughs:
When Manta paths appear as part of the of a url (which they do on most
api requests), reserved characters must be encoded just like in any
other url.

Paths were encoded inconsistently (in that they were sometimes
encoded, sometimes not, and sometimes encoded twice).  When paths were
encoded, spaces were not encoded correctly (turned to `+`) because
despite the tantalizing name `java.net.URLEncoder` does does not
encode strings for inclusion in a url.

The contract for `MantaObject.getPath` has been clarified. It should
always return the "real" (un-encoded) path.

ref #229 #230 231

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/util/MantaUtils.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/util/MantaUtils.java
@@ -7,6 +7,8 @@
  */
 package com.joyent.manta.util;
 
+import io.mikael.urlbuilder.util.Encoder;
+import io.mikael.urlbuilder.util.Decoder;
 import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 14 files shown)


cburroughs:
Please take a look at `MantaClient` in particular, there are many (too many?) different code paths.

java-manta-client/src/main/java/com/joyent/manta/util/MantaUtils.java
```diff
@@ -203,23 +204,27 @@ public static Boolean parseBooleanOrNull(final Object value) {
...
-        final StringBuilder encodedPath = new StringBuilder();
-        for (final String string : elements) {
-            if (string.equals("")) {
-                continue;
-            }
-            encodedPath.append("/").append(URLEncoder.encode(string, "UTF-8"));
-        }
-        return encodedPath.toString();
+    public static String formatPath(final String path) {
+        final Encoder encoder = new Encoder(StandardCharsets.UTF_8);
```
dekobon: [Encoder](https://github.com/mikaelhg/urlbuilder/blob/master/src/main/java/io/mikael/urlbuilder/util/Encoder.java) at first glance seems to be thread safe. Do we want to make static instances of `Encoder` and `Decoder`, so we don't create an instance on every invocation?

cburroughs: Ran unit tests plus `MantaClientPutIT,MantaClientIT,MantaDirectoryListingIteratorIT` with static instances and all seems well.  Will land with that change.

Issue #217, opened 2017-02-23
MantaClient Making Extra HTTP Request For Encrypted Objects


uxcn:
`MantaClient` is making an extra request to manta for unbounded (and negative) byte range requests (specifically `EncryptionHttpHelper`).  `AesCtrCipherDetails#translateByteRange` requires explicit bounds, and has no mechanism to interpret and compute with implicit bounds.

Similarly, implicit encrypted bounds (e.g. `-n`) need to be translated to account for cipher block sizes, cipher block padding, and authentication bytes.  One way (current) is to request the content length from manta.  Per cipher, all the necessary values should already be known though.

Formally, the equations (negative) for `manta_end_pos` should be...

    manta_end_pos = request_end_pos - (block_size - (request_end_pos % block_size)) - mac_auth_bytes (padded cipher)

    manta_end_pos = request_end_pos - mac_auth_bytes (non-padded cipher)

geek:
@uxcn please point to the code that is making the extra request

dekobon:
Hi @uxcn,

I ran [this](https://github.com/joyent/java-manta/blob/3.0/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientRangeIT.java#L125) integration test in encrypted mode. I only say the following requests during the test run:

```
// Adds the test directory
PUT /elijah.zupancic/stor/java-manta-integration-tests/e1db0f2a-3554-426c-8769-43f79c581720 HTTP/1.1
// Adds the test object
PUT /elijah.zupancic/stor/java-manta-integration-tests/e1db0f2a-3554-426c-8769-43f79c581720780cc180-345b-4e9b-8358-79af27f1dbc7 HTTP/1.1
// Gets a range from the object
GET /elijah.zupancic/stor/java-manta-integration-tests/e1db0f2a-3554-426c-8769-43f79c581720780cc180-345b-4e9b-8358-79af27f1dbc7 HTTP/1.1
// Deletes the test data
PUT /elijah.zupancic/stor/java-manta-integration-tests HTTP/1.1
```

Are we missing a test case in our integration tests?

uxcn:
@dekoban The code I was reading sent a HEAD request if I remember correctly, but I might be wrong.  One of the other things I noticed was that there might be an opportunity for optimization (e.g. no mac for range requests); I need to go over the rfc again though.

I'll try debugging this using the integration tests as soon as I have some spare time.

uxcn:
Hey @dekobon,

[...]

PR #237, opened 2017-04-24
MantaClient Making Extra HTTP Request For Encrypted Objects (#217)

uxcn:
This eliminates an extra http (head) request for encrypted byte ranges.


uxcn:
The mockito mocks for stuff don't seem to be being init'd in travis ci for some reason.  I will look into this as soon as I can.

uxcn:
Since this is a fairly large pull request, it probably makes sense to go into a bit more detail.  This pull request eliminates an http head request to a manta server for encrypted object byte ranges, essentially implementing a formal byte range api.

Currently, encrypted byte range requests (relative) require initial round trip tcp bytes from the client to the server, before any actual byte request, to fix offsets that are dependent on the object size (e.g. account for block size and mac); essentially making the request twice.  This unfortunately adds to the request latency, as well as making the request non-atomic (request more likely to fail).

It unfortunately leaves a couple things unaddressed, which were previously unaddressed (e.g. a closed byte range request that accidentally, or possibly intentionally, requests plaintext bytes that are part of the ciphertext mac [not currently tested]).  Prior code assumes no invalid requests, which I think might result in a negative file offset.

The integration tests pass in us-east an on a local machine, but seem to be failing in travis-ci.  I can't determine why, but the testng `@BeforeSuite` annotation doesn't seem to be being parsed.

tjcelaya:
@uxcn Could you explain the algorithm for determining explicit byte ranges based on implicit ranges a bit more? How would we be able to calculate the end of the range if we don't know how big the object is? Your formulas mentioned in #217 neglect to explain the way some of the values are defined, but `request_end_pos` depends on the object size and would need to make the same HEAD request.

```
manta_end_pos = request_end_pos - (block_size - (request_end_pos % block_size)) - mac_auth_bytes (padded cipher)

manta_end_pos = request_end_pos - mac_auth_bytes (non-padded cipher)
```

Quoting your last comment:
> Currently, encrypted byte range requests (relative) require initial round trip tcp bytes from the client to the server, before any actual byte request, to fix offsets that are dependent on the object size (e.g. account for block size and mac); essentially making the request twice.

While there are two requests occurring it is an overgeneralization to say "making the request twice" since the `HEAD` request only contains metadata needed to perform the second request correctly. The block comment above the line you referenced which issues the HEAD request mentions this fact: https://github.com/joyent/java-manta/blob/master/java-manta-client/src/main/java/com/joyent/manta/http/EncryptionHttpHelper.java#L370

Running the integration test suite with `mvn verify` on your branch causes every `MantaClientRangeIT` test to fail with the following exception thrown from your code:

```
(com.joyent.manta.client.MantaClientRangeIT)  Time elapsed: 0.651 sec  <<< FAILURE!
com.joyent.manta.exception.MantaException: incorrect separator, should be ' ' (range=bytes 827-892/893)
	at com.joyent.manta.http.MantaHttpHeaders.getContentRange(MantaHttpHeaders.java:1208)
	at com.joyent.manta.http.EncryptionHttpHelper.httpRequestAsInputStream(EncryptionHttpHelper.java:558)
	at com.joyent.manta.client.MantaClient.getAsInputStream(MantaClient.java:407)
	at com.joyent.manta.client.MantaClientRangeIT.canGetWithUnboundedStartRange(MantaClientRangeIT.java:136)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1128)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
	at org.testng.TestRunner.privateRun(TestRunner.java:782)
	at org.testng.TestRunner.run(TestRunner.java:632)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:366)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:319)
	at org.testng.SuiteRunner.run(SuiteRunner.java:268)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1244)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1169)
	at org.testng.TestNG.run(TestNG.java:1064)
	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:132)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:112)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.execute(TestNGDirectoryTestSuite.java:99)
	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:147)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:290)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:242)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:121)
```

uxcn:
> Could you explain the algorithm for determining explicit byte ranges based on implicit ranges a bit more? How would we be able to calculate the end of the range if we don't know how big the object is? Your formulas mentioned in #217 neglect to explain the way some of the values are defined, but request_end_pos depends on the object size and would need to make the same HEAD request.

The values submitted to the server are entirely dependent on properties of the ciphers.  There are still issues (even in the current implementation) when values requested eat into the `mac`, but those can be caught in the response (assuming manta is rfc-7233 compliant'ish).

> While there are two requests occurring it is an overgeneralization to say "making the request twice" since the HEAD request only contains metadata needed to perform the second request correctly. The block comment near the line you link to explains why the request needs to happen:

The major issue, which might or might not be a problem for users, is that an initial `head` request is another http/network request, which eats into an application request response budget.

For example, assuming a (reasonably generous) 300ms budget for a response...

    <application request> => <head request (t ms)> => <head response (v ms)> => <get request (t ms)> => <application response>

If *t* is ~150ms, the application misses its deadline (user interactivity in this case).  Sans `head` request, there's a fairly roomy 150ms to spare.

It's (hopefully) a less common use case, but the additional request still eats into the user's application response time budget.

In all sincerity, my two big concerns with this pull request were usability and security, which is why I took as extreme caution as I did to identify the potential errors and security flaws (including possible user errors), and move as much possible to compile time (hopefully saving some grief).

uxcn:
> Finally, running the integration test suite with mvn verify on your branch causes every MantaClientRangeIT test to fail with the following exception thrown from your code:

Looking at the exception, the `Content-Range` value doesn't look like it matches the [rfc-7233](https://tools.ietf.org/html/rfc7233) examples although I might be wrong (slightly late).

Issue #242, opened 2017-05-06
Add JCE Policy Key Size Check


uxcn:
The java-manta api currently assumes the unlimited strength [jce policy](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html) is installed, which causes errors at runtime as larger than permitted aes key sizes are requested (e.g. 192/256).

tjcelaya:
What sort of check did you have in mind? It seems like the best we can do is compare the requested key size at config time against [`javax.crypto.Cipher.getMaxAllowedKeyLength`](http://docs.oracle.com/javase/6/docs/api/javax/crypto/Cipher.html#getMaxAllowedKeyLength%28java.lang.String%29) and throw an exception if the requested size is greater than the allowed value. Need to investigate the portability of this solution though.

uxcn:
What I had in mind was moving the error, at very least, earlier in the execution path, and providing a user friendly error message.  In this case, having the jvm catch the error based on the jce policy strength doesn't allow for any unintended consequences, but catching errors earlier can hopefully save people time.

I wrote an example that does this when the associated cipher details are constructed, although it could easily be moved to compile/load time.  The example preserves portability across jvms with and without the unlimited strength jce policy, although there are a number of different approaches possible depending on the needs.

I'm submitting a pull request, although it's a fairly large code, and it potentially changes a public interface people may already be depending on.  Comments are very much welcome.

tjcelaya:
Closed by #244 which was a reengineering of #243. Thanks for helping us make this more explicit!

PR #243, opened 2017-05-08
Add JCE Policy Key Size Check (#242)

uxcn:
This pull request moves jvm [jce policy](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html) errors earlier into the execution path, and gives a user friendly error message when a requested cipher details object exceeds the maximum policy key length.

It's possible to move the error even further in the execution path, to load or compile time, although the example preserves compatibility across jvms with and without an unlimited strength jce policy.

As such, please note this alters the public `SupportedCipherDetails` objects to be instantiated at execution time rather than load or compile time, which may alter part of the api people may already depend on.

Comments are very welcome.

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
@@ -9,6 +9,8 @@ package com.joyent.manta.client.crypto;
 
 import org.apache.commons.lang3.Validate;
 
+import java.security.NoSuchAlgorithmException;
+
 import javax.crypto.Cipher;
 
 /**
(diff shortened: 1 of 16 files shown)


tjcelaya:
I'd like to close this PR in favor of #244, which achieves the desired result while making less changes to the external interface. It also adds test annotations for the tests which rely on stronger encryption so they can optionally be skipped.

The core changes can be seen [here](https://github.com/joyent/java-manta/compare/master...tjcelaya:094abfd) and achieve the same effect as your PR. The rest of the changes in that PR were JavaDoc cleanup and test annotations.

Please discuss issues further in the future, specially when refactoring is involved or interfaces change.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
```diff
@@ -22,24 +24,101 @@
     /**
      * Instance of AES128-CBC cipher.
      */
-    public static final AesCbcCipherDetails INSTANCE_128_BIT = new AesCbcCipherDetails(128);
+    private static volatile AesCbcCipherDetails instance128Bit;
```
tjcelaya: This change would break any consumer which uses `INSTANCE_{128,192,265}_BIT` and as such would be a breaking change.

java-manta-client/src/test/java/com/joyent/manta/client/crypto/AesCbcCipherDetailsTest.java
```diff
@@ -8,86 +8,104 @@
...
 public class AesCbcCipherDetailsTest extends AbstractCipherDetailsTest {
+
+    private AesCbcCipherDetails AES_CBC_128;
+
+    private AesCbcCipherDetails AES_CBC_192;
+
+    private AesCbcCipherDetails AES_CBC_256;
+
+    @BeforeClass
+    private void init() throws NoSuchAlgorithmException {
```
tjcelaya: The fact that this `init` method was defined on every test that interacts with any of the AesCipherDetails classes proves you're aware of the problems caused by removing the `INSTANCE` constants.

uxcn:
> The fact that this init method was defined on every test that interacts with any of the AesCipherDetails classes proves you're aware of the problems caused by removing the INSTANCE constants.

The `init` method isn't actually necessary.  The reason it was added is that, as an example, it caches the lazily initialized instances.

Functionally, there is no difference (performance... slightly).

tjcelaya:
https://github.com/joyent/java-manta/pull/244#issuecomment-300498628

> However, I did take extreme caution to ensure all manta users would be able to use the changes without a negative impact . . . The biggest tangible change was admittedly, and somewhat regrettably, moving something currently public behind functions.

Imagining a situation where the goal is "retrieving the largest possible integer" your statement amounts to "I know everyone wrote `Integer.MAX_VALUE` everywhere in their code but it's ok as long as you go fix all of those to now use `Integer.getMaxValue()`".

We can change the interface if we bump the version, but the changelog for 4.0 would have just two lines:
- Feature: JCE is optional now, unless you want stronger ciphers, then you still need it
- Breaking Change: rewrite your client code whether or not you already had JCE, sorry.

I don't think our users would appreciate that.

tjcelaya:
https://github.com/joyent/java-manta/pull/244#issuecomment-300498628

uxcn:
> Imagining a situation where the goal is "retrieving the largest possible integer" your statement amounts to "I know everyone wrote Integer.MAX_VALUE everywhere in their code but it's ok as long as you go fix all of those to now use Integer.getMaxValue()".

I think the context is maybe a bit different (security and hopefully not the entire jdk userbase), although my honest hope was that no code outside the cilent library was already using the interface since it was only recently released (hoped to get as many comments as possible).

> We can change the interface if we bump the version, but the changelog for 4.0 would have just two lines:

Yeah... I usually consider breaking an api to require a major version change too.  Normally, I try to preserve backward compatibility at all costs, and allow everyone to move forward at their own pace.  There are some exceptions to that, for me at least, though (e.g. something is fundamentally broken, security flaw, prohibitive maintenance cost, and so on...).  This bug is... debatable (less if people are already using the public interface).

> Feature: JCE is optional now, unless you want stronger ciphers, then you still need it

The way I did (or at least tried) this one was that it gives an easy to understand error message when you try to use something that requires the unlimited strength jce policy and it isn't installed (so hopefully it helps any users accomplish their ultimate goal with minimal complication and frustration).  Unless you use it, you generally don't need to read documentation or worry about it.  Requiring it at compile time could be a possibility, but the question I would ask is if doing so poses risks, saves time, and/or frustration (also flexibility).

> Breaking Change: rewrite your client code whether or not you already had JCE, sorry.

If an api break is necessary, it definitely sucks (especially if it's a result of something you don't use), but this would possibly not be an effective way to communicate it.  I would probably explain how it changed and why (e.g. compiling code can result in security flaw).  Otherwise, normally just add the new api and hopefully implement the existing api as much as possible in terms of it.

tjcelaya:
Apologies, PR was reopened by clicking the wrong comment button.

PR #244, opened 2017-05-09
#242 jce size check

tjcelaya:
This PR is based off of PR #243 and addresses issue #242. @uxcn: While we would like to be more explicit about the library's requirements, the same effect can be accomplished without changing the external interface.

Using the [null object pattern](https://en.wikipedia.org/wiki/Null_Object_pattern) to allow for construction of illegal ciphers without changing the public interface. Makes minor adjustments to constructor visibility (`private` is now `protected`, could be package-private). 

Tests which use the stronger ciphers have been marked with the ~`strong-crypto`~ `unlimited-crypto` group so they can be skipped in case installing Java Cryptography Extensions is not desired or possible:

~`mvn test -DexcludedGroups=strong-crypto`~
`mvn test -DexcludedGroups=unlimited-crypto`

### DIFF
--- a/java-manta-client-kryo-serialization/src/test/java/com/joyent/manta/serialization/CipherSerializerTest.java
+++ b/java-manta-client-kryo-serialization/src/test/java/com/joyent/manta/serialization/CipherSerializerTest.java
@@ -50,12 +50,12 @@ public class CipherSerializerTest {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_128_BIT);
     }
 
-    @Test(groups = {"strong-crypto"})
+    @Test(groups = {"unlimited-crypto"})
     public void canSerializeAesGcm192() throws Exception {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_192_BIT);
     }
 
-    @Test(groups = {"strong-crypto"})
+    @Test(groups = {"unlimited-crypto"})
     public void canSerializeAesGcm256() throws Exception {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_256_BIT);
     }
@@ -64,12 +64,12 @@ public class CipherSerializerTest {
         canSerializeCipher(AesCtrCipherDetails.INSTANCE_128_BIT);
     }
 
-    @Test(groups = {"strong-crypto"})
+    @Test(groups = {"unlimited-crypto"})
     public void canSerializeAesCtr192() throws Exception {
         canSerializeCipher(AesCtrCipherDetails.INSTANCE_192_BIT);
     }
 
-    @Test(groups = {"strong-crypto"})
+    @Test(groups = {"unlimited-crypto"})
     public void canSerializeAesCtr256() throws Exception {
         canSerializeCipher(AesCtrCipherDetails.INSTANCE_256_BIT);
     }
@@ -78,12 +78,12 @@ public class CipherSerializerTest {
         canSerializeCipher(AesCbcCipherDetails.INSTANCE_128_BIT);
     }
 
-    @Test(groups = {"strong-crypto"})
+    @Test(groups = {"unlimited-crypto"})
     public void canSerializeAesCbc192() throws Exception {
         canSerializeCipher(AesCbcCipherDetails.INSTANCE_192_BIT);
     }
 
-    @Test(groups = {"strong-crypto"})
+    @Test(groups = {"unlimited-crypto"})
     public void canSerializeAesCbc256() throws Exception {
         canSerializeCipher(AesCbcCipherDetails.INSTANCE_256_BIT);
     }
--- a/java-manta-client-kryo-serialization/src/test/java/com/joyent/manta/serialization/CipherSerializerTest.java
+++ b/java-manta-client-kryo-serialization/src/test/java/com/joyent/manta/serialization/CipherSerializerTest.java
@@ -50,10 +50,12 @@ public class CipherSerializerTest {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_128_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
     public void canSerializeAesGcm192() throws Exception {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_192_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
     public void canSerializeAesGcm256() throws Exception {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_256_BIT);
     }
@@ -62,10 +64,12 @@ public class CipherSerializerTest {
         canSerializeCipher(AesCtrCipherDetails.INSTANCE_128_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
     public void canSerializeAesCtr192() throws Exception {
         canSerializeCipher(AesCtrCipherDetails.INSTANCE_192_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
     public void canSerializeAesCtr256() throws Exception {
         canSerializeCipher(AesCtrCipherDetails.INSTANCE_256_BIT);
     }
@@ -74,10 +78,12 @@ public class CipherSerializerTest {
         canSerializeCipher(AesCbcCipherDetails.INSTANCE_128_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
     public void canSerializeAesCbc192() throws Exception {
         canSerializeCipher(AesCbcCipherDetails.INSTANCE_192_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
     public void canSerializeAesCbc256() throws Exception {
         canSerializeCipher(AesCbcCipherDetails.INSTANCE_256_BIT);
     }
--- a/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
@@ -14,7 +14,6 @@ import java.security.NoSuchAlgorithmException;
  * Class for constructing AesCipherDetails which may or may not be useable in the current runtime.
  *
  * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
- * @since 3.0.0
  */
 public final class AesCipherDetailsFactory {
 
@@ -64,8 +63,6 @@ public final class AesCipherDetailsFactory {
      *
      * @param mode cipher mode
      * @param requestedKeyLengthBits secret key length in bits
-     *
-     * @return static
      */
     static SupportedCipherDetails buildWith(final CipherMode mode, final int requestedKeyLengthBits) {
 
--- a/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
@@ -36,7 +36,7 @@ public final class AesCipherDetailsFactory {
     /**
      * Maximum AES key length when Java Cryptography Extensions are missing.
      */
-    private static final int MAX_KEY_LENGTH_FALLBACK = 128;
+    static final int MAX_KEY_LENGTH_FALLBACK = 128;
 
     /**
      * Maximum key length allowed by current runtime.
--- a/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
@@ -58,6 +58,10 @@ public final class AesCipherDetailsFactory {
     }
 
     /**
+     * Builds a concrete subclass of AbstractAesCipherDetails, or a LocallyIllegalAesCipherDetails when the current
+     * runtime lacks the required <a href="https://en.wikipedia.org/wiki/Java_Cryptography_Extension">Java Cryptography
+     * Extensions</a>.
+     *
      * @param mode cipher mode
      * @param requestedKeyLengthBits secret key length in bits
      *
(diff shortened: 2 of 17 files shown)


java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
```diff
@@ -0,0 +1,94 @@
...
+            // will return MAX_KEY_LENGTH_DEFAULT if JCE missing, catch is for the compiler
+        } catch (NoSuchAlgorithmException nsae) {
+            maxKeyLength = MAX_KEY_LENGTH_DEFAULT;
+        }
+
+        MAX_KEY_LENGTH_ALLOWED = maxKeyLength;
+    }
+
+    /**
+     * @param mode cipher mode
```
dekobon: I would add a description of the method.

tjcelaya: Description added, thanks for catching that

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
```diff
@@ -0,0 +1,98 @@
...
+package com.joyent.manta.client.crypto;
+
+import javax.crypto.Cipher;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * Class for constructing AesCipherDetails which may or may not be useable in the current runtime.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
+ * @since 3.0.0
```
cburroughs: This is a new class, right? I don't think this `@since` can be right?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    @SuppressWarnings("checkstyle:JavadocVariable")
+    public enum CipherMode {
+        CBC,
+        CTR,
+        GCM
+    }
+
+    /**
+     * Default maximum key length.
```
cburroughs: We expect the "normal" thing to be using the unlimited encryption thought, right? I'm not sure "default" is the best term.

java-manta-client-kryo-serialization/src/test/java/com/joyent/manta/serialization/CipherSerializerTest.java
```diff
@@ -50,10 +50,12 @@ public void canSerializeAesGcm128() throws Exception {
         canSerializeCipher(AesGcmCipherDetails.INSTANCE_128_BIT);
     }
 
+    @Test(groups = {"strong-crypto"})
```
cburroughs: I'm a little worried about the implication that "128" isn't "strong" crypto.  Maybe "unlimited" to align with what the policy files are called?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/LocallyIllegalAesCipherDetails.java
```diff
@@ -0,0 +1,159 @@
...
+import org.bouncycastle.crypto.macs.HMac;
+
+import javax.crypto.Cipher;
+import java.security.spec.AlgorithmParameterSpec;
+
+/**
+ * Implementation of an AESCipherDetails that MUST NOT be used within the current runtime.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
+ * @since 3.0.0
```
cburroughs: Same comment re `since`

java-manta-client/src/main/java/com/joyent/manta/client/crypto/LocallyIllegalAesCipherDetails.java
```diff
@@ -0,0 +1,159 @@
...
+     */
+    LocallyIllegalAesCipherDetails(final int keyLengthBits) {
+        this.keyLengthBits = keyLengthBits;
+    }
+
+    /**
+     * Fail immediately, we don't belong on this JVM.
+     */
+    private void fail() {
+        throw new Error("This cipher is not compatible with the current runtime: (keyLengthBits="
```
cburroughs: Can you add a comment for why this is an `Error` and not `Exception`?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Builds a concrete subclass of AbstractAesCipherDetails, or a LocallyIllegalAesCipherDetails when the current
+     * runtime lacks the required <a href="https://en.wikipedia.org/wiki/Java_Cryptography_Extension">Java Cryptography
+     * Extensions</a>.
+     *
+     * @param mode cipher mode
+     * @param requestedKeyLengthBits secret key length in bits
+     *
+     * @return static
```
cburroughs: ?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
tjcelaya: Clarified comment.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
tjcelaya: Apologies, this was copied the block from AbstractAesCipherDetails. Removed.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/AesCipherDetailsFactory.java
tjcelaya: Method and JavaDoc used to be in AbstractAesCipherDetails, `@return` doesn't even make sense on a constructor. Removed.

(`@return static` looks like phpDoc annotation, force of habit I guess...)

java-manta-client-kryo-serialization/src/test/java/com/joyent/manta/serialization/CipherSerializerTest.java
tjcelaya: Good point, I've changed the group name to "unlimited-crypto"

tjcelaya:
- Are any changes to the testng xml needed to account for the new group?

I assumed the group would need to be registered in the `testng.xml` file but since the classes are already annotated as `@Test`s I don't believe any configuration changes are necesary. I verified the now-annotated test cases still run under `mvn test` if no extra flags are passed. Unfortunately Intellij IDEA complains that the group is undefined but it doesn't actually seem to cause any problems. My experience with TestNG specifically is limited though so I might be overlooking something.

- Can you comment on the compatibility impact of changing the classes of the static objects (AesGcmCipherDetails -> SupportedCipherDetails)

The Aes*CipherDetails classes don't define any public methods of their own so changing the `INSTANCE_N` types to an interface _should not_ break consumers. Additionally `SupportedCiphersLookupMap` is a `LookupMap<String, SupportedCipherDetails>` so in some cases consumers are already accessing the CipherDetails instances through that interface.

Review comments addressed:
 - test group name "strong-crypto" -> "unlimited-crypto"
 - remove erroneous `@since` and expand comment about Error vs Exception
 - rename "default" key strength constant to "fallback"

Pending questions: `SupportedCiphersLookupMap` private constructor still includes all ciphers, even the illegal ones, still working on this but suggestions are welcome.

tjcelaya:
Regarding the `SupportedCiphersLookupMap` class: it doesn't really make sense to keep the illegal AesCipherDetails instances in the map so I modified that constructor as well. Even though access to these is more indirect than the `Aes*CipherDetails.INSTANCE_N` it would still fails the same way. 

Please review:
https://github.com/joyent/java-manta/pull/244/files#diff-76c25a6e5828e06fa8c4ebbe0676fd1a

uxcn:
My two biggest concerns with the missing jce policy strength check were that the error happens later in user's software than necessary (which could easily cause unintended consequences), as well as a lack of a clear concise message that helps the user diagnose the issue quickly.

If I am honest, I am not really sure this addresses those concerns since by design the `null object` will error whenever an operation is requested, instead of when the corresponding cipher object is instantiated.

Similarly, some of the code seemed potentially a bit more complex than necessary (e.g. aes cipher type `enum`), possibly giving additional attack surface area, as well as potential maintenance issues leading to exploits further down.

I also noticed a couple other slightly more minor things like `protected` instead of `private` and treating a fatal error (missing aes implementation) as a non-error.    Last, I noticed that some of the variable names seemed possibly confusing or maybe even misleading (e.g. `DEFAULT_MAX_KEY_LENGTH`).  Using `Cipher.getMaxAllowedKeyLength` directly may be a bit clearer..

tjcelaya:
The library installation requirements list JCE alongside Java and Maven.
Despite this fact, it possible to use the Manta client without JCE as long
as your code does not attempt to invoke encryption algorithms which would
_eventually_ throw an error anyways. We can make an effort to accommodate
these use-cases but it needs to be handled carefully to prevent other users
from experiencing issues.

The fact that `public static final` constants are one of the primary
interfaces when invoking encryption does expose the nonportable aspects of
the JCE dependency. Unfortunately, changing the public interface to the
`crypto` package in order to make errors occur slightly more quickly based
on your proposed implementation would cause compiler issues for the users
that satisfied the installation requirements without providing any benefit.
This doesn't seem like a change that most users would be interested in
seeing.

We can plan to revamp the portability of the way we perform encryption but
it would probably require much more discussion and planning.

tjcelaya:
@uxcn:
> Similarly, some of the code seemed potentially a bit more complex than necessary (e.g. aes cipher type enum), possibly giving additional attack surface area, as well as potential maintenance issues leading to exploits further down.

Could you please explain how an enum listing the cipher modes supported by the library compromises security?

uxcn:
@tjcelaya My sincere apologies if I was overly critical.  I was only hoping to offer constructive advice toward hopefully improving the code for users.

> The fact that `public static final` constants are one of the primary interfaces when invoking encryption does expose the nonportable aspects of the JCE dependency. Unfortunately, changing the public interface to the `crypto` package in order to make errors occur slightly more quickly based on your proposed implementation would cause compiler issues for the users that satisfied the installation requirements without providing any benefit.  This doesn't seem like a change that most users would be interested in seeing.

I only offered the example as a possible solution.  However, I did take extreme caution to ensure *all* manta users would be able to use the changes without a negative impact (unlimited strength jce vs. non-unlimited, parallel contention, etc...).  The biggest tangible change was admittedly, and somewhat regrettably, moving something currently public behind functions.

> Could you please explain how an enum listing the cipher modes supported by the library compromises security or increases the probability of a security issue?

I only offered it as a general concern.  Although, the `AesCipherDetails` and `enum` types are disjoint, so the compiler can't enforce a lot of things.  For example, if another `AesCipherDetails` type is added, the compiler can't enforce that any code that depending on the `enum` type exists.  Some possibilities I can think of might be degrading encryption strength, removing cipher authentication, or possibly even crashing the jvm.

tjcelaya:
@uxcn No need to apologize! I'll admit that some of your concerns are valid but, again, the design changes you implemented prohibit us from accepting your submission as-is. Please keep in mind that discussing your proposed solution with us before implementing it *significantly* raises the chances that we can merge your PR directly. The more details we can agree on before any code is written, the easier things become for everyone.

cburroughs:
I think this is looking good, a few things (some of which I should have brought up before, sorry!):
 * Did you get a clean run with all of the integration tests? 
 * Do we need an `unlimited-crypto` test group for integration tests too? My inclination is that the config permutations are already pretty complicated there so it is not worthwhile to support, but I wanted to make sure we had discussed it.
 * I don't think I see a unit test that involves `LocallyIllegalAesCipherDetails`?  I would not advise covering every method, but a small "expect this to throw" check would be reasonable ROI.
 * Could you add a brief line to the CHANGELOG? Something along the lines of "while unlimited strength policy is strongly encouraged... we now handle more gracefully".

tjcelaya:
- Did you get a clean run with all of the integration tests?

Integration tests are timing out on my computer (on all branches), I'm working on getting them set up in Travis CI.

- Do we need an unlimited-crypto test group for integration tests too? My inclination is that the config permutations are already pretty complicated there so it is not worthwhile to support, but I wanted to make sure we had discussed it.

The [testng-it.xml](https://github.com/joyent/java-manta/blob/master/java-manta-it/src/test/resources/testng-it.xml) specifies 128-bit encryption only so I'm inclined to say no. Does "config permutations" refer to the possible combinations of configurations or the test configuration? If that's the only file we need to check then I don't see a reason to add the group to the integration test suite yet.

- I don't think I see a unit test that involves LocallyIllegalAesCipherDetails? I would not advise covering every method, but a small "expect this to throw" check would be reasonable ROI.

Didn't think about this but it's a good idea. Added a LocallyIllegalAesCipherDetailsTest which expects the error-throwing behavior.

- Could you add a brief line to the CHANGELOG? Something along the lines of "while unlimited strength policy is strongly encouraged... we now handle more gracefully".

Added.

Unrelated to those issues, I noticed a line that seems to be out of place while editing the README:

> If you want to skip running of the test suite, use the -DskipTests property.

It occurs right under the example for programmatically setting `manta.user` and `manta.key_id` but seems to belong under the [building From Source](https://github.com/joyent/java-manta#from-source) block about `mvn package` so I moved it up. Let me know if I should revert this change.

cburroughs:
With "permutations" I meant that we run the same test with several different cipher configurations.

I think all of the code looks good.  I'd like to make sure the integration tests "pass" before landing (a few may be flaky still despite cleanup).  Feel free to reach out if you need help looking at anything strange you are seeing.  (Which manta you are integrating against also matters.)  While long I can usually run them from my laptop over wifi.

dekobon:
For what it is worth, I use a server in the east region as a build, release, integration test server. You may want to do that as well because it helps to mitigate timeout issues.

Issue #245, opened 2017-05-23
Range Requests Into And Beyond MAC Are Not Taken Into Account


uxcn:
Users can currently make range requests past the end of the actual ciphertext, which results in the *mac* being treated as ciphertext.

tjcelaya:
I decided to write a simple main method to avoid the complexity of the test suite to verify this issue. Feel free to fork and modify it to fit the situation you're describing and me know if you have any questions, I tried to be as explicit as possible:

https://gist.github.com/tjcelaya/203735c84db3554e4c1796ff67427311

tjcelaya:
@uxcn Apologies I hadn't seen your comment on the code example, I'll test out your suggestion this afternoon.

PR #246, opened 2017-05-23
Demonstrate #245.

uxcn:
Remove additional logic from unit test to more closely match manta behavior, and
add explicit test for seeking past ciphertext.

**please note:** this intentionally breaks tests to demonstrate issue

### DIFF
--- a/java-manta-client/src/test/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStreamTest.java
+++ b/java-manta-client/src/test/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStreamTest.java
@@ -502,6 +502,12 @@ public class MantaEncryptedObjectInputStreamTest {
         canReadByteRangeAllReadModes(cipherDetails, startPos, endPos);
     }
 
+    public void canReadByteRangeEndBytesNonOffsetAesCtr128() throws IOException {
+        SupportedCipherDetails cipherDetails = AesCtrCipherDetails.INSTANCE_128_BIT;
+        int endPos = this.plaintextSize + cipherDetails.getAuthenticationTagOrHmacLengthInBytes() - 1;
+        canReadByteRangeAllReadModes(cipherDetails, 0, endPos);
+    }
+
 
     public void canCopyStreamWithLargeBufferBufferAuthenticatedAesCbc128() throws IOException {
          canCopyToOutputStreamWithLargeBuffer(AesCbcCipherDetails.INSTANCE_128_BIT, true);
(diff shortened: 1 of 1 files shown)


java-manta-client/src/test/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStreamTest.java
```diff
@@ -932,9 +931,6 @@ private void canReadByteRange(SupportedCipherDetails cipherDetails,
              * the binary of the actual object (even if the range went beyond). */
             long ciphertextByteRangeLength = Math.min(ciphertextRangeMaxBytes, ciphertextRangeRequestBytes);
 
-            if (unboundedEnd)
-                ciphertextByteRangeLength = ciphertextSize - ciphertextStart; // include MAC
```
tjcelaya: This line seems to be attempting to prevent the situation you're describing, though `unboundedEnd` might not be the best name since it'll be true when `endPositionExclusive >= plaintextSize` (i.e. explicitly requesting an end that is beyond the plaintext size like the situation you describe in #245 )

tjcelaya:
Regarding the test case you added `canReadByteRangeEndBytesNonOffsetAesCtr128`, could you refactor it to not use `canReadByteRangeAllReadModes` and only include the code needed to demonstrate the unexpected data being returned? That helper method calls `canReadByteRange` which makes it difficult to diagnose exactly what is going wrong and where.

java-manta-client/src/test/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStreamTest.java
uxcn: That's kind of part of my point; the test has additional logic that either overlaps with other classes or is missing.

I remember reading `ENCRYPTION_PLAINTEXT_CONTENT_LENGTH` in the original specification; reading through the code though, the impression I got was that it wasn't a value that could be *guaranteed* to be accurate.

If you check the branch the byte-range pull request was based on, the implementation I chose was to depend on rfc values, since the *manta* server should be able to compute object length atomically, and plaintext/ciphertext boundaries are pure functions of the ciphers and object length.  `Content-Range` object length is not *required* per rfc-7233... although worst case fallback could still be making an additional `head` request to *manta*.

uxcn:
@tjcelaya I'll take a look at the code later tonight.

I appreciate you taking the time to look into this for me.

tjcelaya:
Circled back to the discussion in the [linked gist](https://gist.github.com/tjcelaya/203735c84db3554e4c1796ff67427311). If I'm understanding correctly, the bug can be reproduced by `put`ting an object and then forcing an incorrect value into the `m-encrypt-plaintext-content-length` field. The central finding is that need to be more explicit about which headers users are allowed to set. I'm opening a new issue to update `MantaClient.ILLEGAL_METADATA_HEADERS` and potentially start throwing exceptions when users set metadata fields that we intend to overwrite.

Issue #248, opened 2017-06-05
getAsInputStream throws exceptions for empty files


tjcelaya:
When requesting a file from Manta using `MantaClient.getAsInputStream`, an exception is thrown in cases where the file is empty (newlines inserted for clarity):

```
java.lang.UnsupportedOperationException:
We are expecting an instance of [class org.apache.http.conn.EofSensorInputStream] as a stream, 
but instead received: [class org.apache.http.impl.io.EmptyInputStream]
  at com.joyent.manta.http.StandardHttpHelper.lambda$httpRequestAsInputStream$0(StandardHttpHelper.java:306)
  at com.joyent.manta.http.StandardHttpHelper.executeRequest(StandardHttpHelper.java:505)
  at com.joyent.manta.http.StandardHttpHelper.httpRequestAsInputStream(StandardHttpHelper.java:330)
  at com.joyent.manta.client.MantaClient.getAsInputStream(MantaClient.java:407)
  at com.joyent.manta.client.MantaClient.getAsInputStream(MantaClient.java:446)
  at com.joyent.manta.client.MantaClientIT.canReadEmptyFileAsInputStream(MantaClientIT.java:177)
```

It seems trivial to return a `new NullInputStream(0)` since we are already using Commons IO. PR with failing example test coming soon.

PR #249, opened 2017-06-05
getAsInputStream test for empty file, issue #248

tjcelaya:
Failing test case for issue #248

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
@@ -167,6 +167,22 @@ public class MantaClientIT {
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


cburroughs:
Is this still a WIP until we decide on the approach?

tjcelaya:
Yes it's just been neglected in as a result of the newer issues coming up. EofSensorInputStream plays a big role there so I was considering a dummy MantaObjectInputStream implementation to avoid changing the class that actually has to deal with real data.

Also wondered how empty files interact with encryption since there are assumptions about plaintext/ciphertext and authentication data.

In summary, yes I'm intending to come back to this. If you have any suggestions or warnings I'd be glad to hear them out.

tjcelaya:
Resolved by #409

Issue #252, opened 2017-06-13
Creating SnapLinks with + in their path fail


dekobon:

cburroughs:
:-(  Do you have a stack trace?

dekobon:
My tests linked in this issue were broken. It turns out that this isn't the root cause of the user reported issue for the Manta Hadoop FS driver.

PR #253, opened 2017-06-13
Add SnapLink tests and additional exception context for SnapLink failures

dekobon:
This PR adds some minor improvements to tests and exception messages for SnapLinks. These changes were spawned out of the investigation of a bug reporded with the Manta Hadoop Filesystem driver.

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
@@ -347,11 +347,11 @@ public class MantaClientIT {
     @Test
     public final void testPutLinkWithPlusInPath() throws IOException {
         final String name = UUID.randomUUID().toString();
-        final String prefix = testPathPrefix + MantaClient.SEPARATOR + "+tmp";
+        final String destPrefix = testPathPrefix + MantaClient.SEPARATOR + "+tmp";
+        mantaClient.putDirectory(destPrefix);
 
-        mantaClient.putDirectory(prefix);
         final String sourcePath = testPathPrefix + name;
-        final String destPath = prefix + MantaClient.SEPARATOR + name;
+        final String destPath = destPrefix + MantaClient.SEPARATOR + name;
 
         mantaClient.put(sourcePath, TEST_DATA);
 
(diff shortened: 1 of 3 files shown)


cburroughs:
If #252 has turned out to be a bug elsewhere, can this be closed?

dekobon:
#252 turned out to not be a bug, but I do think we should take these changes. @cburroughs thoughts?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -482,6 +512,7 @@ public final void testList() throws IOException {
     }
 
     @Test(expectedExceptions = MantaObjectException.class)
+    @SuppressWarnings("ReturnValueIgnored")
```
cburroughs: Is this related?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -494,6 +525,7 @@ public final void testListNotADir() throws IOException {
     }
 
     @Test(expectedExceptions = MantaClientHttpResponseException.class)
+    @SuppressWarnings("ReturnValueIgnored")
```
cburroughs: Is this related?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
dekobon: I had to add these annotations to get the library to compile on my platform.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
dekobon: I had to add these annotations to get the library to compile on my platform.

dekobon:
> I think for most of the existing funky character tests I just did ~!@#$%^&*()_+ all at once. If you prefer to break it down into smaller tests I understand the tradeoff, but I'm not sure I see a reason to test an arbitrary smaller set of troublesome characters for snaplinks than for other api calls.

The tests that matter are the tests that verify that the encoding works for the location header because it uses a different code path on the server and we want to verify that deals with encoded characters correctly.

cburroughs:
I understand the need the tests for `+=@`. My question is why not also have tests for `~!#$%^&*()` or the troublesome space if it is a different code path.

dekobon:
> My question is why not also have tests for ~!#$%^&*() or the troublesome space if it is a different code path.

Good point. I'll update the PR to reflect that.

PR #258, opened 2017-06-14
Add support for MantaClient.move() to recursively create destination directories

### LINKED ISSUE #256: MantaClient.move() does not support moving a file to a non-existent directory
Current move() behavior does not let you move a file into a directory that does not currently exist. However, because S3 and Swift both virtualize directories in the form of keys, many libraries (I'm looking at you Hadoop Filesystem) support move semantics that allow for creating a directory structure when moving a file.

dekobon:
Fixes #256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project aims to adhere to [Semantic Versioning](http://semver.org/).
 
+## [3.1.1] - 2017-?
+### Fixed
+ - [MantaClient.move() does not support moving a file to a non-existent directory](https://github.com/joyent/java-manta/issues/256)
+
 ## [3.1.0] - 2017-06-07
 
 Several related URL encoding bugs have been fixed.  Objects with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project aims to adhere to [Semantic Versioning](http://semver.org/).
 ## [3.1.1] - 2017-?
 ### Fixed
  - [MantaClient.move() does not support moving a file to a non-existent directory](https://github.com/joyent/java-manta/issues/256)
+   This fix optionally allows users to recursively create a destination directory 
+   structure on move(). This is useful for supporting third-party APIs/libraries
+   that use this pattern due to a S3 first design. 
 
 ## [3.1.0] - 2017-06-07
 
--- a/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -1496,12 +1496,15 @@ public class MantaClient implements AutoCloseable {
             throws IOException {
         LOG.debug("Moving [{}] to [{}]", source, destination);
 
-        MantaObjectResponse entry = head(source);
+        final String formattedSource = formatPath(source);
+        final String formattedDestination = formatPath(destination);
+
+        MantaObjectResponse entry = head(formattedSource);
 
         if (entry.isDirectory()) {
-            moveDirectory(source, destination, entry);
+            moveDirectory(formattedSource, formattedDestination, entry);
         } else {
-            moveFile(source, destination);
+            moveFile(formattedSource, formattedDestination);
         }
     }
 
--- a/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -1517,10 +1517,10 @@ public class MantaClient implements AutoCloseable {
             throws IOException {
         LOG.debug("Moving [{}] to [{}]", source, destination);
 
+        MantaObjectResponse entry = head(source);
+
         final String formattedSource = formatPath(source);
         final String formattedDestination = formatPath(destination);
-
-        MantaObjectResponse entry = head(formattedSource);
 
         if (entry.isDirectory()) {
             moveDirectory(formattedSource, formattedDestination, entry);
--- a/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -1524,13 +1524,10 @@ public class MantaClient implements AutoCloseable {
 
         MantaObjectResponse entry = head(source);
 
-        final String formattedSource = formatPath(source);
-        final String formattedDestination = formatPath(destination);
-
         if (entry.isDirectory()) {
-            moveDirectory(formattedSource, formattedDestination, entry);
+            moveDirectory(source, destination, entry);
         } else {
-            moveFile(formattedSource, formattedDestination, recursivelyCreateDestinationDirectories);
+            moveFile(source, destination, recursivelyCreateDestinationDirectories);
         }
     }
 
@@ -1550,7 +1547,8 @@ public class MantaClient implements AutoCloseable {
     private void moveFile(final String source, final String destination,
                           final boolean recursivelyCreateDestinationDirectories)
             throws IOException {
-        final String destinationDir = FilenameUtils.getFullPath(destination);
+        final String formattedDestination = formatPath(destination);
+        final String destinationDir = FilenameUtils.getFullPath(formattedDestination);
 
         if (recursivelyCreateDestinationDirectories && !existsAndIsAccessible(destinationDir)) {
                 putDirectory(destinationDir, true);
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project aims to adhere to [Semantic Versioning](http://semver.org/).
 
+## [3.1.1] - 2017-?
+### Fixed
```
cburroughs: You labeled the ticket an enhancement :-)  I think it would be good to include the brief S3 compatibility reasoning in the hopes of being maximally helpful to anyone actually using manta on it's own terms.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1478,9 +1479,14 @@ public void putSnapLink(final String rawLinkPath, final String rawObjectPath,
     }
 
     /**
-     * Moves a file from one path to another path. This operation is not
-     * transactional and will fail or produce inconsistent result if the source
-     * or the destination is modified while the operation is in progress.
+     * <p>Moves a file or directory from one path to another path. When moving
```
cburroughs: "file or directory" --> object?

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1478,9 +1479,14 @@ public void putSnapLink(final String rawLinkPath, final String rawObjectPath,
...
+     * directories or files between different directories, this operation is
+     * not transactional and will fail or produce inconsistent result if
```
cburroughs: "may" fail right? We can't guarantee to detect a concurrent modification.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1493,46 +1499,80 @@ public void move(final String source, final String destination)
...
+     * the source or the destination is modified while the operation is
+     * in progress.
+     *
+     * @param source Original path to move from
+     * @param destination Destination path to move to
+     *
+     * @throws IOException thrown when something goes wrong
+     */
+    private void moveFile(final String source, final String destination) throws IOException {
+        final String destinationDir = FilenameUtils.getFullPath(destination);
```
cburroughs: Is this going to do the right thing on a path where `formatPath` has not been called yet?

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1493,46 +1499,80 @@ public void move(final String source, final String destination)
...
+                String relPath = sourcePath.substring(sourceDirPath.length());
+                String destFullPath = destDirPath + SEPARATOR + relPath;
+
+                if (mantaObject.isDirectory()) {
+                    move(mantaObject.getPath(), destFullPath);
+                } else {
+                    putSnapLink(destFullPath, sourcePath, null);
+                }
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
```
cburroughs: Doesn't this method already throw a checked IOException? Which I guess was the existing behavior :-/

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -400,8 +426,7 @@ public final void canMoveEmptyDirectory() throws IOException {
             + path);
     }
 
-
-    public void moveDirectoryWithContents(final String source, final String destination) throws IOException {
+    private void moveDirectoryWithContents(final String source, final String destination) throws IOException {
```
cburroughs: Why the visibility change?  Should this be in the new `move` group?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -374,7 +374,33 @@ public final void canMoveFile() throws IOException {
...
+    }
+
+    @Test(groups = "move")
+    public final void canMoveFileToDifferentDirectory() throws IOException {
+        final String name = UUID.randomUUID().toString();
+        final String source = testPathPrefix + name;
+        mantaClient.put(source, TEST_DATA);
+
+        final String destination = testPathPrefix +
+                "stor/knoxmetrics-dev-marc/topics/+tmp/knox-metrics/partition=0/4ab7bccf-3f31-4b5c-bc74-35926084fd3d_tmp.parquet";
```
cburroughs: The "knoxmetrics-dev-marc" might want to be scrubbed to something generic since it *sounds* like a product name.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: Fixed

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: Fixed

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: Fixed

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
dekobon: Removed this test

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
dekobon: This was originally a mistake to code this method as public. In fact, it was being invoked by the test runner and then subsequently skipped.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: This was a mistake in the past implementation. I believe that I fixed this in other parts of the code previously. By just throwing an `UncheckedIOException` it simplifies debugging.

CHANGELOG.md
dekobon: Updated changelog.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1478,61 +1479,127 @@ public void putSnapLink(final String rawLinkPath, final String rawObjectPath,
...
+     *
+     * <p><em>Note: If you need to do a rename, then use
+     * {@link #putSnapLink(String, String, MantaHttpHeaders)}</em></p>
+     *
+     * @param source Original path to move from
+     * @param destination Destination path to move to
+     * @param recursivelyCreateDestinationDirectories when true create the full destination directory path
+     * @throws IOException thrown when something goes wrong
+     */
+    public void move(final String source, final String destination,
```
cburroughs: Is there a reason these parameters shouldn't be Validated? I think that is a pre-existing omission and I'm fine with a separate change.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1478,61 +1479,127 @@ public void putSnapLink(final String rawLinkPath, final String rawObjectPath,
...
+                     final boolean recursivelyCreateDestinationDirectories)
+            throws IOException {
         LOG.debug("Moving [{}] to [{}]", source, destination);
 
-        MantaObjectResponse entry = head(source);
+        final String formattedSource = formatPath(source);
+        final String formattedDestination = formatPath(destination);
```
cburroughs: urlencoding is sadness and I'm afraid my previous comment was misleading.  Since urlencoding is not idempotent it must be done only once.   For example,  if we encode in `move()` the path will be double encoded on calls to `putDirectory`, resulting in erroneous object names.  When I did the last round of encoding changes I tried to differentiate between the `path` (encoded) and `rawPath` (not-encoded) as parameter names, but I didn't go to change methods that were not broken and I'm not sure how much that clarifies the situation.   This is all terribly error prone and maybe it might be better to have a `MantaPath` object if we redo the api yet again.

Anyway, what I was worried about was some corner case where `FilenameUtils.getFullPath` would have surprising results relative to the later encoded path.  I don't think there is one but am not familiar with that code.  Sorry for leading you astray!

I did notice while looking at the docs again that `FilenameUtils.getFullPath` can return `null`.  I'm not sure what they mean by an invalid path, but based on our recent dealings with exceptions-lacking-context we might want to check for that to error in a more useful way.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
cburroughs: Should this be in the new move group?

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: I added validation to those to parameters and to the head() method.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
dekobon: I rejiggered this so that we don't do two formatPath() calls.

PR #264, opened 2017-06-17
Fix potential file leak from MPU upload part

tjcelaya:
A user was uploading data with client-side encryption enabled through `uploadPart()` and experiencing connection resets. This would result in an exception being thrown that included the message `Caused by: com.joyent.manta.org.apache.http.client.NonRepeatableRequestException: Cannot retry request with a non-repeatable request entity`.

When attempting to pass in a `File` object instead of an `InputStream` it was discovered that files were not being released and the disk in use for temporary storage would become full. The modification made in this PR to `EncryptingPartEntity` was reported to resolve the issue.

Inspecting the non-MPU code path revealed the possibility that the issue could occur there as well. I am awaiting confirmation that `put(String, File)` causes the same problem. Still need to add tests to verify this behavior, though I've been having trouble just reproducing it.

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
@@ -33,6 +33,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.FileInputStream;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
--- a/java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
@@ -13,12 +13,13 @@ import com.joyent.manta.client.MantaObjectInputStream;
 import com.joyent.manta.config.ConfigContext;
 import com.joyent.manta.config.IntegrationTestConfigContext;
 import com.joyent.manta.exception.MantaClientException;
-import com.joyent.manta.exception.MantaClientHttpResponseException;
-import com.joyent.manta.exception.MantaErrorCode;
 import com.joyent.manta.exception.MantaMultipartException;
 import com.joyent.manta.http.MantaHttpHeaders;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.RandomUtils;
+import org.apache.http.entity.FileEntity;
+import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testng.Assert;
@@ -30,12 +31,14 @@ import org.testng.annotations.Parameters;
 import org.testng.annotations.Test;
 
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.concurrent.Executors;
 import java.util.stream.Stream;
 
 import static org.testng.Assert.assertEquals;
@@ -52,6 +55,8 @@ public class EncryptedServerSideMultipartManagerIT {
 
     private static final int FIVE_MB = 5242880;
 
+    private static final String TEST_FILENAME = "Master-Yoda.jpg";
+
     private String testPathPrefix;
 
     @BeforeClass()
@@ -65,13 +70,18 @@ public class EncryptedServerSideMultipartManagerIT {
         }
         mantaClient = new MantaClient(config);
 
+        // sanity check
+        if (!mantaClient.existsAndIsAccessible(config.getMantaHomeDirectory())) {
+            Assert.fail("Invalid credentials, cannot proceed with test suite");
+        }
+
         if (!mantaClient.existsAndIsAccessible(config.getMantaHomeDirectory()
                 + MantaClient.SEPARATOR + "uploads")) {
             throw new SkipException("Server side uploads aren't supported in this Manta version");
         }
 
         multipart = new EncryptedServerSideMultipartManager(this.mantaClient);
-        testPathPrefix = String.format("%s/stor/java-manta-integration-tests/%s",
+        testPathPrefix = String.format("%s/stor/java-manta-integration-tests/%s/",
                 config.getMantaHomeDirectory(), UUID.randomUUID());
         mantaClient.putDirectory(testPathPrefix, true);
     }
@@ -325,7 +335,6 @@ public class EncryptedServerSideMultipartManagerIT {
         }
     }
 
-
     public final void canUploadWithSinglePartAndPerformRangeRequest() throws IOException {
         final String name = UUID.randomUUID().toString();
         final String path = testPathPrefix + name;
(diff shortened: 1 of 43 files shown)


cburroughs:
Was the  `NonRepeatableRequestException` the result of the leak?  The cause and effect is unclear to me.

As I understand the leak problem, the input streams be passed to `uploadPart(UPLOAD upload, int partNumber, InputStream inputStream)` are never being closed.  That could be a problem for any stream associated with external resources, and was noticed in the case of `FileInputStream` where the file remains open.  Is that correct?  If so, I think an integration test could verify the fix by uploading parts made of temporary files and verifying that they are deleted.

>Still need to add tests to verify this behavior, though I've been having trouble just reproducing it.

Were you able to reproduce it?

dekobon:
I might be able to add some color to this PR. When an InputStream is used as a source to a put() method and if there is an error, there is no way to rewind the stream and restart. Thus, @tjcelaya suggested that they use a put() with a File object instead. When they were doing this, he discovered an issue.
The issue was that the FileInputStream that was opened for the passed File object was not close correctly when the upload completed or aborted. @tjcelaya Is that a correct summary?

tjcelaya:
@cburroughs In my testing I was able to delete the file successfully despite leaving an open `FileInputStream`. Confusingly, I _can still read bytes_ out of the supposedly-deleted file. When running a simple test that terminates quickly I'm led to believe the actual file handle will be released and the space reclaimed once the JVM terminates, but in a long-running process the space on-disk would silently pile up.

This is still a really confusing issue to me but the user confirmed they experienced the issue with regular `put(String, File)` and that the second edit I made fixed that file leak as well. My next idea is to create a ridiculously small `tmpfs` partition and push tempfiles through it to validate this claim.

tjcelaya:
So, I've successfully reproduced the issue! Unfortunately I have no idea how to write a test around this.

Using [this script](https://gist.github.com/Roman2K/3238fb441e298369198e) to mount a 35MB tmpfs-like volume at a folder referenced by `DIR_TINY`, I can repeatedly run the following test and get the same result. A zero-byte file will be left behind when the second copy fails and the test crashes.

```
    public void testFilesAreReleasedProperly() throws Exception {
        // copying to a 35MB ramdisk mounted at DIR_TINY
        File f1 = new File(DIR_TINY, randomUUIDString());
        File f2 = new File(DIR_TINY, randomUUIDString());

        // FILE_ORIGINAL is ~30MB
        System.out.println("copy: " + f1.getAbsolutePath());
        FileUtils.copyFile(FILE_ORIGINAL, f1);
        System.out.println("ok");

        FileInputStream f1i = new FileInputStream(f1);
        ByteArrayOutputStream baos = new ByteArrayOutputStream();
        IOUtils.copy(f1i, baos);                // read bytes from f1 but don't close it
        Assert.assertTrue(f1.delete());         // delete successfully

        System.out.println("copy: " + f2.getAbsolutePath());
        FileUtils.copyFile(FILE_ORIGINAL, f2);  // java.io.IOException: No space left on device
        throw new Error("shouldn't make it here");
    }
```

Output:

```
copy: /Users/tomascelaya/sandbox/tiny/42128b1e-45cc-4323-a91f-f80c87f205a3
ok
copy: /Users/tomascelaya/sandbox/tiny/e25570ad-cdf3-4582-b2a0-3a256bf30f12

java.io.IOException: No space left on device

	at sun.nio.ch.FileDispatcherImpl.pwrite0(Native Method)
	at sun.nio.ch.FileDispatcherImpl.pwrite(FileDispatcherImpl.java:66)
	at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:89)
	at sun.nio.ch.IOUtil.write(IOUtil.java:51)
	at sun.nio.ch.FileChannelImpl.writeInternal(FileChannelImpl.java:778)
	at sun.nio.ch.FileChannelImpl.write(FileChannelImpl.java:764)
	at sun.nio.ch.FileChannelImpl.transferFromFileChannel(FileChannelImpl.java:636)
	at sun.nio.ch.FileChannelImpl.transferFrom(FileChannelImpl.java:708)
	at org.apache.commons.io.FileUtils.doCopyFile(FileUtils.java:1151)
	at org.apache.commons.io.FileUtils.copyFile(FileUtils.java:1088)
	at org.apache.commons.io.FileUtils.copyFile(FileUtils.java:1040)
	at com.joyent.manta.client.MantaWriteInputStreamIT.testFilesAreReleasedProperly(MantaWriteInputStreamIT.java:96)
```

Any advice for constraining this to a simple test case is welcome.

cburroughs:
I think what you are observing with both tmpfs and the original issue is that even if a file is unlinked, it still exists (and you can read and write to it!) as long as there is at least one open handle.

```python
import os
import os.path

f = open('myfile.txt', 'w+')
f.write('a\n')
f.write('b\n')
print 'file exists:', os.path.exists(f.name)
print 'unlinking file...'
os.unlink(f.name)
print 'file exists:', os.path.exists(f.name)
f.write('c\n')
f.seek(0, 0)
print f.read()
f.close()
```

results in 
```
file exists: True
unlinking file...
file exists: False
a
b
c

```

Since the file exists, the OS can not reclaim any space.  This can be particularly hard to debug since `ls` and friends won't show it.  The fix is to -- as you did in this diff -- close the file object when you are done with it.  I think this matches everything you described, but maybe the smaller case feels a bit less spooky?

I believe this change can be tested by:
 * Adding a test to `EncryptedServerSideMultipartManagerIT` where we pass in a `FileInputStream`.  (This is a code path that should be covered regardless.)
 * After the upload, assert that `theInputStream.getFD().valid()` is false (indicating a closed and thus invalid fd). See [FileDescriptor](https://docs.oracle.com/javase/7/docs/api/java/io/FileDescriptor.html)

java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
```diff
@@ -65,13 +70,18 @@ public void beforeClass(@org.testng.annotations.Optional Boolean usingEncryption
...
+        }
+
         if (!mantaClient.existsAndIsAccessible(config.getMantaHomeDirectory()
                 + MantaClient.SEPARATOR + "uploads")) {
             throw new SkipException("Server side uploads aren't supported in this Manta version");
         }
 
         multipart = new EncryptedServerSideMultipartManager(this.mantaClient);
-        testPathPrefix = String.format("%s/stor/java-manta-integration-tests/%s",
+        testPathPrefix = String.format("%s/stor/java-manta-integration-tests/%s/",
```
tjcelaya: The missing slash meant files were created just outside the test directory with the test directory name as the first half of their filename. This is part of the reason test data wasn't being completely cleaned up after test runs.

java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
```diff
@@ -65,13 +70,18 @@ public void beforeClass(@org.testng.annotations.Optional Boolean usingEncryption
         }
         mantaClient = new MantaClient(config);
 
+        // sanity check
```
tjcelaya: Verify that user credentials are valid by checking for existence of home folder. This will prevent us from skipping the test suite because we assume MPU is not available on the Manta instance used for running the tests.

tjcelaya:
Travis-CI is currently failing because of the following error:

`(argument mismatch; org.apache.http.HttpEntity cannot be converted to com.joyent.manta.org.apache.http.HttpEntity)`

I need to figure out what's going on with shading that's preventing the use of that class in the `java-manta-it` project. The same class shows up in a test in `java-manta-client` so I'm looking at the shade configs to see if there's something missing from the IT project's pom.

tjcelaya:
`mvn verify` succeeded at [2017-06-24T01:08:52+00:00](https://ci.tjcelaya.co/blue/rest/organizations/jenkins/pipelines/java-manta%20verify%20PR%20264/runs/11/log/?start=0)

Proceeding with merge.

PR #271, opened 2017-06-26
reduce durability test level: 4-->3

cburroughs:
A smaller number lowers the burden for staging/test manta instances to
support a passing integration test run.


cburroughs:
@dekobon I was not sure if there was anything that was originally special about "4".

dekobon:
There was no specific reason to choose four. The rational was choosing a non-default value would be better for the tests.

PR #275, opened 2017-06-28
Fix MD5 mismatch request ID missing

tjcelaya:
Found another exception that's missing a request ID. Not sure if `HttpHelper.annotateContextedException` might be overkill in this case.


dekobon:
To summarize my understanding of this change: you are adding context including a request id for checksum validation failure exceptions?

tjcelaya:
Your summary is correct.

Issue #278, opened 2017-06-29
ServerSideMultipartManager.complete exceptons missing context


cburroughs:
in `ServerSideMultipartManager.complete`, if an exception occurs while issuing the `commit` HTTP POST, the exception is not wrapped in additional context.  For example:

```
java.net.SocketTimeoutException: Read timed out
        at java.net.SocketInputStream.socketRead0(Native Method)
        at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
        at java.net.SocketInputStream.read(SocketInputStream.java:171)
        at java.net.SocketInputStream.read(SocketInputStream.java:141)
        at sun.security.ssl.InputRecord.readFully(InputRecord.java:465)
        at sun.security.ssl.InputRecord.read(InputRecord.java:503)
        at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:973)
        at sun.security.ssl.SSLSocketImpl.readDataRecord(SSLSocketImpl.java:930)
        at sun.security.ssl.AppInputStream.read(AppInputStream.java:105)
        at com.joyent.manta.org.apache.http.impl.io.SessionInputBufferImpl.streamRead(SessionInputBufferImpl.java:137)
        at com.joyent.manta.org.apache.http.impl.io.SessionInputBufferImpl.fillBuffer(SessionInputBufferImpl.java:153)
        at com.joyent.manta.org.apache.http.impl.io.SessionInputBufferImpl.readLine(SessionInputBufferImpl.java:282)
        at com.joyent.manta.org.apache.http.impl.io.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:129)
        at com.joyent.manta.org.apache.http.impl.io.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:53)
        at com.joyent.manta.org.apache.http.impl.io.AbstractMessageParser.parse(AbstractMessageParser.java:259)
        at com.joyent.manta.org.apache.http.impl.DefaultBHttpClientConnection.receiveResponseHeader(DefaultBHttpClientConnection.java:163)
        at com.joyent.manta.org.apache.http.impl.conn.CPoolProxy.receiveResponseHeader(CPoolProxy.java:165)
        at com.joyent.manta.org.apache.http.protocol.HttpRequestExecutor.doReceiveResponse(HttpRequestExecutor.java:273)
[...]

PR #279, opened 2017-06-29
capture context for exceptions thrown during MPU commit request

cburroughs:
ref #278

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/client/multipart/ServerSideMultipartManager.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/multipart/ServerSideMultipartManager.java
@@ -618,6 +618,14 @@ public class ServerSideMultipartManager extends AbstractMultipartManager
                 LOGGER.info("Multipart upload [{}] for file [{}] from [{}] parts has completed",
                             upload.getId(), location.getValue(), numParts);
             }
+        } catch (MantaMultipartException | MantaClientHttpResponseException me) {
+            // Already annotated
+            throw me;
+        } catch (Exception e) {
+            String msg = "Error initiating multipart upload completion";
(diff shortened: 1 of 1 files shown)


cburroughs:
This ends up double wrapping during  validation failures, breaking an integration tests. :-(  Working on a revised fix.

cburroughs:
Let me know if there are any concerns about the revised diff.

tjcelaya:
Is there a reason for those two exception types specifically? Not a blocker just wondering why you didn't use more general types like `catch (MantaException | MantaIOException me)`.

dekobon:
They are separate because their inheritance is separate. A `MantaIOException` can be caught by any `catch (IOException e)` clause. Whereas a more generalized MantaException doesn't have to do with IO and a consumer may want to handle all exceptions relating to Manta in a single try/catch clause. What is difficult is that Java doesn't support multiple inheritance and if I remember right you can't catch an interface.

tjcelaya:
**TL;DR** style question was too terse, sounded like a language feature question

I'm aware of the inheritance difference, that's why I combine them in the proposed single-catch. I would assume `catch (MantaException | MantaIOException me) { throw me; }` would achieve the intended effect (rethrowing already-annotated exceptions) and was curious why more specific exception types were selected and had separate catch clauses.

Stepping back to think about it a new kind of exception could be thrown from that `try` and my proposed change would assume that exception is annotated correctly, which might not be the case. This change is more explicit (though the `MantaMultipartException` and `MantaClientHttpResponseException` clauses could still be combined since they do the same thing, but that's more about style).

dekobon:
Fair point.

cburroughs:
Switched to `|`, I'm still forget about the 'new' multiple catch blocks.

PR #280, opened 2017-06-29
Fix MD5 mismatch during retry

tjcelaya:
Since we allow retries for most sensible cases it is possible that a HttpEntity is used to write output more than once. `DigestedEntity` accepts a `digest` object which keeps internal state that was not being cleaned up between calls to `writeTo`. Some manual testing with a proxy revealed that `HttpEntity` objects are reused as-is when performing a retry. This can lead to an actually-successful retry failing on the client when comparing the client-calculated MD5 with the server's MD5 response header. This will manifest itself in an exception like the following:

```
Caused by: com.joyent.manta.exception.MantaChecksumFailedException: Client calculated MD5 and server calculated MD5 do not match
Exception Context:
        [1:mantaSdkVersion=3.1.2]
        [2:serverMd5=29d7cf482158399d1d83dddf44af1511]
        [3:clientMd5=e85d394ceb64b5e745bad680705ee1c7]
---------------------------------
        at com.joyent.manta.http.StandardHttpHelper.validateChecksum(StandardHttpHelper.java:363)
        at com.joyent.manta.http.StandardHttpHelper.httpPut(StandardHttpHelper.java:222)
        at com.joyent.manta.http.EncryptionHttpHelper.httpPut(EncryptionHttpHelper.java:217)
        at com.joyent.manta.client.MantaClient.put(MantaClient.java:1218)
        at com.joyent.manta.client.MantaClient.put(MantaClient.java:1163)
```

Verified that there are no changes that need to be made for `EncryptingEntity` since it doesn't have any of its own internal state and builds a new `CipherOutputStream` at the beginning of `writeTo`.

### DIFF
--- a/java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
+++ b/java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
@@ -1,0 +1,50 @@
+package com.joyent.manta.http.entity;
+
+import com.twmacinta.util.FastMD5Digest;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+
+/**
+ * Created by tomascelaya on 6/29/17.
+ */
+@Test
+public class DigestedEntityTest {
+
+    public void testWriteToProducesReliableDigest() throws Exception {
+        String content = RandomStringUtils.randomAlphanumeric(100);
+        DigestedEntity entity = new DigestedEntity(
+                new ExposedStringEntity(content, StandardCharsets.UTF_8),
+                new FastMD5Digest());
+
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        entity.writeTo(out);
+        byte[] initialDigest = entity.getDigest();
+
+        // connection reset, httpclient performs a retry reusing the same entity
+        out.reset();
+        entity.writeTo(out);
+        byte[] retryDigest = entity.getDigest();
+
+        Assert.assertTrue(
+                Arrays.equals(initialDigest, retryDigest),
+                "Reuse of DigestedEntity produced differing digest for first retry");
+
+
+        // connection reset again
+        out.reset();
+        entity.writeTo(out);
+        byte[] extraRetryDigest = entity.getDigest();
+
+        Assert.assertTrue(
+                Arrays.equals(initialDigest, extraRetryDigest),
+                "Reuse of DigestedEntity produced differing digest for second retry");
+    }
+
+}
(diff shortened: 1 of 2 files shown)


java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
```diff
@@ -0,0 +1,50 @@
+package com.joyent.manta.http.entity;
```
cburroughs: ^^ Need headers.

java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
```diff
@@ -0,0 +1,50 @@
...
+        out.reset();
+        entity.writeTo(out);
+        byte[] extraRetryDigest = entity.getDigest();
+
+        Assert.assertTrue(
+                Arrays.equals(initialDigest, extraRetryDigest),
+                "Reuse of DigestedEntity produced differing digest for second retry");
+    }
+
+}
```
cburroughs: nit: I am surprised checkstyle let the missing newline by.

java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
```diff
@@ -0,0 +1,50 @@
...
+        byte[] initialDigest = entity.getDigest();
+
+        // connection reset, httpclient performs a retry reusing the same entity
+        out.reset();
+        entity.writeTo(out);
+        byte[] retryDigest = entity.getDigest();
+
+        Assert.assertTrue(
+                Arrays.equals(initialDigest, retryDigest),
+                "Reuse of DigestedEntity produced differing digest for first retry");
```
cburroughs: Isn't this asserting that the digests are the *same*, that they do not differ?

java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+
+        // connection reset again
+        out.reset();
+        entity.writeTo(out);
+        byte[] extraRetryDigest = entity.getDigest();
+
+        Assert.assertTrue(
+                Arrays.equals(initialDigest, extraRetryDigest),
+                "Reuse of DigestedEntity produced differing digest for second retry");
```
cburroughs: ^^

java-manta-client/src/test/java/com/joyent/manta/http/entity/DigestedEntityTest.java
tjcelaya: The message is what is shown when the test case fails, output like the following:

```
java.lang.AssertionError: Reuse of DigestedEntity produced differing digest expected [true] but found [false]
Expected :true
Actual   :false
```

Unfortunately using `assertEquals` on the digests directly produces odd output, but using `assertTrue` without a message would be even less helpful.

tjcelaya:
@cburroughs Addressed questions, please re-review

Issue #276, opened 2017-06-28
MantaClientPutIT#testPutWithJPGFile failures on LX ubuntu instance


cburroughs:
On an instance like so:

```
# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.2 LTS"
# uname -a
Linux cf5b8960-613e-c7e2-f6cb-97c615bc8a06 4.3.0 BrandZ virtual linux x86_64 x86_64 x86_64 GNU/Linux
# java -version
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-0ubuntu1.16.04.2-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
```

I am seeing consistent failures at:
```
(com.joyent.manta.client.MantaClientPutIT)  Time elapsed: 0.177 sec  <<< FAILURE!
java.lang.AssertionError: Content type wasn't detected correctly expected [image/jpeg] but found [text/plain]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:513)
        at org.testng.Assert.assertEqualsImpl(Assert.java:135)
        at org.testng.Assert.assertEquals(Assert.java:116)
        at org.testng.Assert.assertEquals(Assert.java:190)
        at com.joyent.manta.client.MantaClientPutIT.testPutWithJPGFile(MantaClientPutIT.java:333)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
        at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
        at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
        at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1128)
        at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
[...]

PR #281, opened 2017-07-05
adjust mimetype checks to reduce platform specifc behavior

cburroughs:
ref #276

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/http/ContentTypeLookup.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/http/ContentTypeLookup.java
@@ -74,12 +74,23 @@ public final class ContentTypeLookup {
         String type =  ObjectUtils.firstNonNull(
                 // Use explicitly set headers if available
                 headerContentType,
-                // Probe using the JVM default detection method
-                Files.probeContentType(file.toPath()),
-                // Detect based on destination filename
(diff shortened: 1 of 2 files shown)


dekobon:
This is a sensible change. The logic behind the original ordering was that `Files.probeContentType()` was a more accurate detector of mime types than a filename. However, when implemented the platform specific peculiarities weren't taken into account. Both choices are compromises. The correct answer is that the user brings their own mime-type.

tjcelaya:
Since the issue seems to stem from `Files.probeContentType(Path)` being platform-specific, I'm inclined to ask: would it be too much of a breaking change to stop using it entirely? We're already altering functionality with this change by favoring the more consistent detection method, it seems like an opportunity to drop the inconsistent method entirely and have more predictable behavior.

dekobon:
If we drop the mimetype autodetect out of the SDK, we will need to implement it upstack in each library that implements the SDK, like VFS and the Hadoop FS driver. That's ok and arguably we should do it there. I look at this feature as useful fluff. It is nice to see the correct mimetype appear in Manta when you have dragged a file into a box in a third-party component.

dekobon:
Let's merge!

dekobon:
Dang, I just remembered that we didn't update the changelog. Can you do that?

Issue #282, opened 2017-07-07
spurious zero byte object failures on EncryptedServerSideMultipartManagerIT.cantUploadSmallMultipartString


cburroughs:
I have been tracking down test failures along these lines.  I have not been able to reproduce them consistently or find a pattern to when they occur.

```
(com.joyent.manta.client.multipart.EncryptedServerSideMultipartManagerIT)  Time elapsed: 3.926 sec  <<< FAILURE!
com.joyent.manta.exception.MantaMultipartException: 
Missing part of multipart upload
Exception Context:
        [1:mantaSdkVersion=3.1.3-SNAPSHOT]
        [2:missing_part=2]
---------------------------------
        at com.joyent.manta.client.multipart.AbstractMultipartManager.lambda$validateThatThereAreSequentialPartNumbers$0(AbstractMultipartManager.java:61)
        at java.util.stream.ReduceOps$1ReducingSink.accept(ReduceOps.java:80)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.ArrayList.forEach(ArrayList.java:1249)
        at java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:390)
        at java.util.stream.Sink$ChainedReference.end(Sink.java:258)
        at java.util.stream.Sink$ChainedReference.end(Sink.java:258)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.reduce(ReferencePipeline.java:474)
        at com.joyent.manta.client.multipart.AbstractMultipartManager.validateThatThereAreSequentialPartNumbers(AbstractMultipartManager.java:59)
        at com.joyent.manta.client.multipart.EncryptedMultipartManager.validateThatThereAreSequentialPartNumbers(EncryptedMultipartManager.java:345)
        at com.joyent.manta.client.multipart.EncryptedServerSideMultipartManagerIT.cantUploadSmallMultipartString(EncryptedServerSideMultipartManagerIT.java:453)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #283, opened 2017-07-07
remove parts on server validation in cantUploadSmallMultipartString

cburroughs:
ref #282

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
@@ -444,10 +444,10 @@ public class EncryptedServerSideMultipartManagerIT {
             assert exception.getCause() instanceof IllegalStateException;
         }
         Assert.assertEquals(uploadedParts.size(), 1, "only first small upload should succeed");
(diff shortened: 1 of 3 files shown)


java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
```diff
@@ -444,10 +444,10 @@ public void cantUploadSmallMultipartString() throws IOException {
             assert exception.getCause() instanceof IllegalStateException;
```
tjcelaya: Should this be a test assertion or use the `assert` keyword?

cburroughs: Well that is the `assert` keyword right?  One of the test methods would almost certainly clearer.  I think there might have been some testng vs junit confusion at the time of writing, but either way I don't see why it could not be cleaned up now.

dekobon: I wouldn't use `assert` in unit tests. In some projects, I use asserts in the main code base because they aren't enabled by default. In particular for performance sensitive code paths. However, nowadays, I just tend to make them explicit checks rather than assertions.

Anyways, for this case, I would definitely use a test assertion.

PR #286, opened 2017-07-13
Fix #284 MPU Automatic Retry

### LINKED ISSUE #284: Automatic retries during MPU with CSE breaks files
We've discovered an issue where automatic retries triggered by the Apache HTTP Client library ruins the state of the `CipherOutputStream` and `HmacOutputStream` objects used when resending the same part. Fixing this would require snapshotting the state of the `EncryptionState` object in `EncryptedMultipartUpload` and only updating the state of encryption and digest objects if the part is uploaded successfully. Attached screenshot shows how a manually terminated request is followed by an automatic retry which corrupts the file.

screen shot 2017-07-11 at 10 48 47 am
(click to zoom)

tjcelaya:
For #284:
 - [x] ~Rename `ReflectionUtils` to `MantaReflectionUtils` and move from `java-manta-client-kryo serialization` module to `java-manta-client` module~
 - [x] make `EncryptedMultipartManager` increment `lastPartNumber` _only_ if the part was uploaded successfully
 - [x] create `HMacCloner` and tests
 - [x] create `CipherCloner` and tests
 - [x] create `EncryptionStateRecorder`
 - [x] make `EncryptedMultipartManager` use `EncryptionStateCloner` to checkpoint encryption state between parts
 - [x] verify `EncryptedServerSideMultipartManagerIT#canRetryUploadPart` passes
 - [x] ~test everything with and without `libnss` enabled~ determined that memoization of PKCS11 Cipher state is not feasible, all MPU operations must use BouncyCastle
 - [x] modify `EncryptedServerSideMultipartManagerIT#canRetryUploadPart` to include retry for multiple parts, including the first part
 - [x] verify `HmacCloner` works with `-Dcom.twmacinta.util.MD5.NATIVE_LIB_FILE=`


tjcelaya:
Closing this PR until it's actually ready for review.

tjcelaya:
Reopening PR for visibility into ongoing work.

tjcelaya:
PR is ready for review.

java-manta-client-kryo-serialization/pom.xml
```diff
@@ -101,6 +102,10 @@
                     <groupId>com.esotericsoftware</groupId>
                     <artifactId>minlog</artifactId>
                 </exclusion>
+                <exclusion>
```
dekobon: Why have you excluded the objenesis library? I was under the impression that kryo needed it to run.

dekobon: Nevermind. The answer is below.

java-manta-client/pom.xml
```diff
@@ -170,6 +170,20 @@
             <version>${dependency.urlbuilder.version}</version>
         </dependency>
 
+        <!-- object cloning library for memoization of encryption state -->
```
dekobon: Do we want to shade and relocate these libraries or have the main library take it as a direct dependency?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/SupportedCipherDetails.java
```diff
@@ -69,6 +69,14 @@
     Cipher getCipher();
 
     /**
+     * In some cases we <em>need</em> the Cipher to live entirely in Java.
```
dekobon: I don't feel great about this change, but I don't see a better way. @cburroughs may have some suggestions.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/SupportedHmacsLookupMap.java
```diff
@@ -37,15 +37,27 @@
     public static final SupportedHmacsLookupMap INSTANCE = new SupportedHmacsLookupMap();
 
     /**
+     * Enum of supported HMac names.
+     */
+    @SuppressWarnings("checkstyle:JavadocVariable")
+    public enum Name {
```
dekobon: I won't name anything `Name`. Maybe `HmacName`?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/SupportedHmacsLookupMap.java
```diff
@@ -80,6 +93,7 @@ private SupportedHmacsLookupMap() {
      * Finds the HMAC implementation name based on the passed object.
      *
      * @param hmac instance to find name for
```
dekobon: Delete line break.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/SupportedHmacsLookupMap.java
```diff
@@ -54,6 +66,7 @@ private SupportedHmacsLookupMap() {
      * new HMAC instance for every value of this map.
      *
      * @param algorithm algorithm to instantiate HMAC instance as
```
dekobon: Delete line break.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartUpload.java
```diff
@@ -83,6 +89,19 @@ EncryptionState getEncryptionState() {
     }
 
     /**
```
dekobon: Give a _hint_ (not an explanation) the javadoc about why you are doing this and why it is package private.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionStateRecorder.java
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Copyright should be 2017.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionStateRecorder.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.OutputStream;
+import java.lang.reflect.Field;
+
+/**
+ * Helper class for recording state of encryption and authentication state during in-progress MPU uploads.
+ * Used to enable retry of {@link EncryptedMultipartManager#uploadPart} in case of network failure.
+ * Like the <a href="https://sourcemaking.com/design_patterns/memento">Memento Pattern</a> but with less respect
+ * for encapsulation.
+ */
+public class EncryptionStateRecorder {
```
dekobon: This class should have more comments on why we would even need a encryption state recorder.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionStateRecorder.java
```diff
@@ -0,0 +1,176 @@
...
+     * {@link Cipher} cloning helper object.
+     */
+    private static final Cloner<Cipher> CLONER_CIPHER = new CipherCloner();
+
+    /**
+     * Construct an EncryptionStateRecorder which can be used to rewind the state of the given {@link EncryptionState}.
+     *
+     * @param encryptionState object to extract state from (and potentially write back to)
+     */
+    public EncryptionStateRecorder(final EncryptionState encryptionState) {
```
dekobon: The class design here seems oddly inverted. I would expect a single `EncryptionStateRecorder` instance that I would leave laying around. It would then take a `EncryptionState` parameter on `record()` and then return a state object. Then `rewind()` would parse that state object and give me a new `EncryptionState` object.

Help convince me as the reader that your mode of operation is better and why in the comments and code.

java-manta-client/src/main/java/com/joyent/manta/exception/MantaMemoizationException.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/main/java/com/joyent/manta/exception/MantaReflectionException.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/main/java/com/joyent/manta/util/CipherCloner.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/main/java/com/joyent/manta/util/CipherCloner.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Utility class for cloning Cipher objects.
+ */
+public final class CipherCloner implements Cloner<Cipher> {
+
+    /**
+     * Instance of object-cloning library that does deep generic cloning.
+     */
+    private static final com.rits.cloning.Cloner INSTANCE = new com.rits.cloning.Cloner();
```
dekobon: Why are you using the full package name?

java-manta-client/src/main/java/com/joyent/manta/util/Cloner.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/main/java/com/joyent/manta/util/DigestCloner.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/main/java/com/joyent/manta/util/DigestCloner.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public final class DigestCloner implements Cloner<Digest> {
+
+    @Override
+    public Digest createClone(final Digest source) {
+        if (source instanceof Memoable) {
+            return (Digest) ((Memoable) source).copy();
+        }
+
+        throw new NotImplementedException("Clone not implemented for type: " + source.getClass().getCanonicalName());
```
dekobon: Did you forget to finish this or is this intentional. Maybe a comment may be useful.

java-manta-client/src/main/java/com/joyent/manta/util/HmacCloner.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/test/java/com/joyent/manta/util/CipherClonerTest.java
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/test/java/com/joyent/manta/util/HmacClonerTest.java
```diff
@@ -0,0 +1,191 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-it/src/test/java/com/joyent/test/util/FailingInputStream.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
```
dekobon: Change copyright to 2017.

java-manta-client/src/main/java/com/joyent/manta/util/CipherCloner.java
tjcelaya: The `Cloner<T>` interface exists in the same package so the symbol clashes. That  interface was originally an abstract class that behaved similarly to `AbstractSerializer` but the limited need for reflection meant it's less useful now. We could drop the mostly-useless interface or rename it (`Copier` springs to mind for some odd reason).

java-manta-client/pom.xml
tjcelaya: There is a version incompatibility between the newest release of `kryo` (which depends on 2.1) and the available versions of `uk.com.robust-it:cloning`:

kryo-4.0.0 depends on objenesis-2.1
kryo-master depends on objenesis-2.5.1 (though it is uncertain when the next release of kryo will occur)

cloning-1.9.5 depends on objenesis-2.5.1
cloning-1.9.4 depends on objenesis-2.2

The [release notes](https://objenesis.org/notes.html) for objenisis don't seem to imply any breaking changes between 2.1 and 2.5.1 so the issue is specifically related to version numbers.

I used a separate sandbox project depending on java-manta 3.1.4-SNAPSHOT to check the impact of not shading and am confused by the resulting dependency tree which doesn't mention objenesis at all:

```
mvn clean dependency:tree -Dverbose
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building j 1.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO]
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ j ---
[INFO]
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ j ---
[INFO] co.tjcelaya.sandbox:j:jar:1.0-SNAPSHOT
[INFO] +- com.joyent.manta:java-manta-client:jar:3.1.4-SNAPSHOT:compile
[INFO] |  +- org.bouncycastle:bcpkix-jdk15on:jar:1.56:compile
[INFO] |  |  \- (org.bouncycastle:bcprov-jdk15on:jar:1.56:compile - omitted for duplicate)
[INFO] |  +- org.bouncycastle:bcprov-jdk15on:jar:1.56:compile
[INFO] |  +- com.squareup.jnagmp:bouncycastle-rsa:jar:2.0.0:compile
[INFO] |  |  \- com.squareup.jnagmp:jnagmp:jar:2.0.0:compile
[INFO] |  |     \- net.java.dev.jna:jna:jar:4.0.0:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.21:compile
[INFO] +- com.joyent.manta:java-manta-client-kryo-serialization:jar:3.1.4-SNAPSHOT:compile
[INFO] |  +- (com.joyent.manta:java-manta-client:jar:3.1.4-SNAPSHOT:compile - omitted for duplicate)
[INFO] |  +- com.esotericsoftware:kryo:jar:4.0.0:compile
[INFO] |  |  \- com.esotericsoftware:reflectasm:jar:1.11.3:compile
[INFO] |  |     \- org.ow2.asm:asm:jar:5.0.4:compile
[INFO] |  +- com.dorkbox:MinLog-SLF4J:jar:1.8:compile
[INFO] |  \- (org.slf4j:slf4j-api:jar:1.7.21:compile - omitted for duplicate)
[INFO] +- org.slf4j:slf4j-simple:jar:1.7.25:compile
[INFO] |  \- (org.slf4j:slf4j-api:jar:1.7.25:compile - omitted for conflict with 1.7.21)
[INFO] +- commons-cli:commons-cli:jar:LATEST:compile
[INFO] +- commons-io:commons-io:jar:2.5:compile
[INFO] \- org.apache.commons:commons-lang3:jar:3.4:compile```

Comment (cburroughs) [inline java-manta-client/pom.xml]: Can we include a "golden file" to deserialize as part of the kyro tests?  That way  we can be sure that version X objects can still be read if we bump the library dep to X+1.

Regarding shading/relocating.  I think the stance that java-manta has taken so far is to do so wherever possible, so I would start with that as the default position.

File: java-manta-client/pom.xml [hunk 22]
```diff
@@ -170,6 +170,20 @@
...
+        <dependency>
+            <groupId>uk.com.robust-it</groupId>
```
cburroughs: There are a few different deep cloneing libraries, any reason why this one in particular?

The readme says "Also cloning of files and streams might make the JVM crash." Are Files or Streams among the things we are cloning?

java-manta-client/src/main/java/com/joyent/manta/client/crypto/EncryptingPartEntity.java
```diff
@@ -89,7 +89,7 @@ public EncryptingPartEntity(final OutputStream cipherStream,
 
     @Override
     public boolean isRepeatable() {
-        return wrapped.isRepeatable();
+        return false;
```
cburroughs: What does `isRepeatable` actually do?  If we never "repeat" how is the retry and snapshot code triggered?

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartManager.java
```diff
@@ -52,6 +52,12 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EncryptedMultipartManager.class);
 
     /**
+     * Object used to reset the state of encryption/authentication in case
+     * a part needs to be reuploaded.
+     */
+    private static final EncryptionStateRecorder RECORDER = new EncryptionStateRecorder();
```
cburroughs: Why `static`?

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartManager.java
```diff
@@ -277,9 +285,15 @@ public ByteArrayOutputStream call(final long uploadedBytes) throws IOException {
                             }
                         }
                     });
+            final MantaMultipartUploadPart part = wrapped.uploadPart(upload.getWrapped(), partNumber, entity);
             encryptionState.setLastPartNumber(partNumber);
-            return wrapped.uploadPart(upload.getWrapped(), partNumber, entity);
+            return part;
         } finally {
+            if (encryptionState.getLastPartNumber() != partNumber && snapshot != null) {
```
cburroughs: Is the rewind done for *any* error during `wrapped.uploadPart`?  Previously the part ordering validation effectively kept consumers from shooting themselves in the foot with retrying parts.  I'm trying to figure out of this opes up any new error paths.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartManager.java
```diff
@@ -277,9 +285,15 @@ public ByteArrayOutputStream call(final long uploadedBytes) throws IOException {
...
+                // a snapshot was prepared but we didn't make it past uploadPart, rewind EncryptionState
+                RECORDER.rewind(encryptionState, snapshot);
```
cburroughs: Can this possibly throw an exception?  I'm worried about anything happening in the same `finally` as the lock release.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionStateRecorder.java
```diff
@@ -0,0 +1,154 @@
...
+        final CipherOutputStream cipherStream;
+        if (usesHmac) {
+            Validate.notNull(snapshot.getHmac(), "Snapshot hmac must not be null");
+
+            final HmacOutputStream digestStream = ensureHmacWrapsCipherStream(encryptionState.getCipherStream());
+
+            try {
+                writeField(FIELD_HMACOUTPUTSTREAM_HMAC, digestStream, snapshot.getHmac());
+            } catch (IllegalAccessException e) {
+                throw new MantaReflectionException("Failed to overwrite HmacOutputStream's hmac", e);
```
cburroughs: Is there any additional context or IDs from EncryptionState that we can add to this and the following exceptions?

java-manta-it/src/test/java/com/joyent/test/util/FailingInputStream.java
```diff
@@ -0,0 +1,77 @@
...
+        int nextByte = wrapped.read();
+        if (nextByte != EOF) {
+            count.incrementAndGet();
+        }
+        return nextByte;
+    }
+
+    private void failAfterMinimum(final int next) throws SpuriousIOException {
+        if (count.get() + next > minimumBytes) {
+            throw new SpuriousIOException("Random error");
```
cburroughs: nit: This is and arbitrary but not "random" error.  It might be nice to include which byte the error was induced at in the message.

tjcelaya:
# Problem
The current design of MPU operations involves creating a new `Cipher` object and, in cases where the desired algorithm does not provide authentication, a new `HMac` object. These objects are then bound to a `CipherOutputStream` and an `HmacOutputStream` which are used to perform encryption and authentication respectively. Any data written to these streams (during a call to `uploadPart`) changes their state. In the case of a network failure, the state of these objects will be altered by the partially-uploaded content. The diagram below illustrates this situation, the first half of part2's data would be considered part of the plaintext when an automatic retry is triggered:

```
  x = network failure

  Normal operation:
  
   iv part1 part2
  |-- ----- -----|

  Automatic Retry:

   iv part1 part2 part2
  |-- ----- --x   -----|
```

# Proposal
We can remedy this situation by taking snapshots of `Cipher` and `Hmac` states between parts and restoring those snapshots if we encounter an exception during `uploadPart`. Snapshots are recorded by cloning the `Cipher` and `Hmac` objects which sit inside of their respective `OutputStream`s, both of which are extracted and written to using reflection. We expect the following stream hierarchy (the cloned objects are bold):

```
  EncryptionState
    OutputStream cipherStream:
      HmacOutputStream
        HMac <---
        OutputStream out:
          CipherOutputStream extends FilterOutputStream (BouncyCastle)
            Cipher <---
            OutputStream out:
              CloneShieldOutputStream (used to allow writing the HMAC after the ciphertext)
                OutputStream out: =>
                  MultipartOutputStream (used to allow attaching to http socket output)

```
# Limitations
Unfortunately the implementation of `EncryptionStateRecorder` is heavily coupled to the way `EncryptionState`'s `cipherStream` field is constructed but I don't see a better way to accomplish this without adding package-private methods to `EncryptionState`, `EncryptionContext`, `HmacOutputStream` and `CipherOutputStream`, the last of which is actually provided by an external library and would need to be reimplemented. There was some investigation into creating ProxyCipher and ProxyHmac objects but that didn't seem feasible. I've added warnings to `makeCipherOutputForStream` about the coupling with `EncryptionStateRecorder`.

By cloning _only_ the `Cipher` and `HMac` objects we can avoid worrying about extracting and replacing streams and focus on copying and swapping in only the objects that hold necessary state.

java-manta-client/src/main/java/com/joyent/manta/client/crypto/EncryptingPartEntity.java
tjcelaya: The underlying Apache HTTP Client library uses this method to determine if it can automatically retry a request. `MantaHttpRequestRetryHandler` is only responsible for determining whether a request should be retried based on the exception thrown. The `HttpEntity` also gets to decide whether it is eligible for automatic retries by returning a value here. [See HttpClient source for more context](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/execchain/RetryExec.java#L106)

Having this be true or delegate to the wrapped instance means that the HttpClient can invoke `writeTo` on the `HttpEntity` for certain kinds of uploads and not others when in fact it should never automatically retry an upload that could affect `Cipher` state

java-manta-client/pom.xml
tjcelaya: I couldn't find any other libraries for deep-cloning besides this one and the copy functionality provided by kryo. Utilizing kryo would effectively mean bringing `java-manta-client-kryo-serialization` into the main module.

I've limited usage of this library to only cloning `Cipher` objects produced by the BouncyCastle library. The `HmacCloner` and `DigestCloner` classes in this PR leverage the `Memoable` interface and some reflection to clone `Hmac` objects instead.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartManager.java
tjcelaya: `EncryptionStateRecorder` is a helper class and has no state of its own. This reference has been removed and the nature of the class has been clarified by newer commits (private constructor, only static methods)

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartManager.java
tjcelaya: Moved into an inner `try` block to address this.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptedMultipartManager.java
tjcelaya: Using an inner `try` now but yes I'm hesitant to classify exceptions by their likelihood of having affected `Cipher` state an only rewind in certain cases, especially given the narrow scope of the inner `try` block. It seems much more straightforward to _always_ reset encryption state if we don't make it to the `setLastPartNumber` invocation.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/EncryptionStateRecorder.java
tjcelaya: Changed `rewind` to accept the `EncryptedMultipartUpload` object instead to it can retrieve the ID in addition to the part number.

java-manta-client/pom.xml
tjcelaya: I don't think it's within the scope of this change to change the behavior of serialization, in fact I've opted to shade these dependencies specifically to avoid conflict with kryo. That said it makes sense to include a file with serialized objects in it so I'll create an issue to figure out what the content would be and how the tests would be designed.

tjcelaya:
Answering general questions from @cburroughs

- I know the changes are spread across a few different classes but somewhere there should be a Big Theory Statement with a few paragraphs of prose explaining how this fits together.
  -  Added a comment to this PR which outlines the problem, proposed solution, and limitations
- Eventually we will need a CHANGELOG entry. It might be helpful to write that now, to explain how this interacts with NSS and what cases we are committing to work (and equally important, which will not work).
  - Added fix to CHANGELOG in this PR
 - What do we need to do either with tests here or on the CI side to make sure this continues to work in the future.
    - I added a testcase `EncryptedServerSideMultipartManagerIT#canRetryUploadPart` which exercises this behavior withint the context of a full MPU. I'm put together a test which exercises just `EncryptionStateRecorder` today, but its usefulness outside the context of `uploadPart` is dubious
 - Do any of the existing microbenchmarks cover an interaction that includes record? if so, is there a change? if not, some should be added to get a sense of how expensive an operation this is.
    - None of the benchmarks address the modified code, I can try to add a test case today which gives us an idea about this. I used `jvisualvm` to profile the cloning process and found the most expensive portion was the cloning of the `Provider` class referenced by `Cipher`. Using `dontCloneInstanceOf` cut the time taken for cloning from ~250ms to ~15ms but it makes sense to automate this measurement
- I believe some of the integration make strong assumptions about the handling of validated part numbers. Do they all still pass so far?
   - The integration test suite passed when I ran it this morning, though I've made a few changes today. The change in part number is actually used as the indication of whether or not the call to `uploadPart` was successful. The inner `try` block which was introduced as a result of your comments is now very tightly scoped to only surround `uploadPart` and `setLastPartNumber`.

tjcelaya:
@cburroughs @dekobon Ready for a second round of reviews. I would specifically appreciate review of the way the objenisis and cloning libraries are shaded.

PR #291, opened 2017-07-27
Fix #290

### LINKED ISSUE #290: MPU encryption corruption still occurs if response is 503
We've discovered that the MPU encryption issue (#284) will still occur if the server response with a 503 instead of completely aborting the connection. The code path for all requests, even successful ones, checks with the configured `ServiceUnavailableRetryStrategy` to determine whether or not it should retry the request. [Here's the offending line](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/DefaultServiceUnavailableRetryStrategy.java#L73).

I'm going to implement `retryRequest` in `MantaServiceUnavailableRetryStrategy` so that we can prevent an automatic retry if the request context indicates that the request was part of an encrypted MPU upload.

tjcelaya:
Some notes:

- We were not calling `validateStatusCode` in `uploadPart`.. It might be worth adding a suite of unit tests for asserting that exceptions are thrown in the case of invalid response codes.
- The test case exercises status code validation with `assertThrows` but we might not want to throw `MantaMultipartException` so I can create a new `Exception` type if desired.
- The test case does not _exactly_ cover the situation reported (cipher state mutation by inadvertent retry) it merely tests that an invalid response from the server triggers an exception since that's what we depend on in order to call `EncryptionStateRecorder#rewind`
- There might be a better way to indicate that the request should not be retried than using `HttpContext`? I'm open to suggestions for other ways to handle this.

For context please see #290. 

[Line that checks if we should retry the request (triggered for successful requests too)](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/execchain/ServiceUnavailableRetryExec.java#L88)

[Line that was allowing the automatic retry (we were not overriding this in MantaServiceUnavailableRetryStrategy](
https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/DefaultServiceUnavailableRetryStrategy.java#L73)


cburroughs:
Jenkins test again

cburroughs:
> Why did the tests fail on travis but not jenkins?

Answer from talking with @tjcelaya , the failure isn't in `test`, but a later part of the build.  I've changed the jenkins job to do `clean test package` and it now fails too.

tjcelaya:
>Isn't the change to uploadPart changing the signature used by existing downsteram code? I don't think that is a deal breaker given the user base, but it needs to be called out and versioned clearly. Why can't we maintain the existing signature with something that passes null for the context?

My first inclination was to add a second method as you described but `uploadPart(UPLOAD, int, HttpEntity)` is package-private or protected so it isn't actually exposed to users. Defining an extra method in each implementation which passed `null` seemed unnecessary but I'm curious if it would simplify the diff to do it this way.

> I'm unsure I understand the change to retryRequest. I thought the intent of #286 was to allow MPU+CSE retries. But the logic here looks like to always return false for isEncryptedMPUPart?

#286 was concerned with disabling _automatic_ retries for MPU parts. My initial intentions for that bugfix did aim to work with automatic retries but the coordination of retry handlers with mpu managers was less than ideal. With the current workflow users are expected to retry a call of `uploadPart` if an exception is thrown.

> I think we could separate concerns a tiny bit better by having the flag set by the manager be along the lines of "IsRetriable" instead of the retry handler needing to have MPU+CSE specific logic. It would be nice if we didn't need to expose a public field that was irrelevant to consumers of the library.

I'll find a better place and name for the "seriously, don't retry this request" flag.

tjcelaya:
HttpContext attribute for disabling retries moved to `MantaServiceUnavailableRetryStrategy` and made more generic, unfortunately it needs to be public because the classes are in different packages.

Not defining `uploadPart(UPLOAD, int, HttpEntity)` since it would create unneeded duplication and the method was never available for users.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/JobsMultipartManager.java
```diff
@@ -296,7 +297,8 @@ public JobsMultipartUpload initiateUpload(final String path,
     @Override
     MantaMultipartUploadPart uploadPart(final JobsMultipartUpload upload,
                                         final int partNumber,
-                                        final HttpEntity entity) throws IOException {
+                                        final HttpEntity entity,
+                                        final HttpContext context) throws IOException {
```
cburroughs: Can `context` be `null` here? (`entity` is validated)

java-manta-client/src/main/java/com/joyent/manta/client/multipart/ServerSideMultipartManager.java
```diff
@@ -309,7 +310,8 @@ public ServerSideMultipartUpload initiateUpload(final String path,
     @Override
     MantaMultipartUploadPart uploadPart(final ServerSideMultipartUpload upload,
                                         final int partNumber,
-                                        final HttpEntity entity)
+                                        final HttpEntity entity,
+                                        final HttpContext context)
```
cburroughs: Same question re `null'.

java-manta-client/src/main/java/com/joyent/manta/client/multipart/AbstractMultipartManager.java
```diff
@@ -197,7 +198,8 @@ public PART uploadPart(final UPLOAD upload,
      */
     abstract PART uploadPart(UPLOAD upload,
                              int partNumber,
-                             HttpEntity entity) throws IOException;
+                             HttpEntity entity,
+                             HttpContext context) throws IOException;
```
cburroughs: Missing javadoc on `context`?  I had some nullability questions below, but they might be better addressed with prose here.

tjcelaya: Forgot the JavaDoc, thanks for calling it out. Regarding the nullability of `HttpContext` it's fine to pass null since the only param that is required is `HttpUriRequest`. `ServersideMultipartManagerIT#canUploadWithSinglePartByteArray` validates that this is the case. See [CloseableHttpClient's source for reference](canUploadWithSinglePartByteArray)

PR #293, opened 2017-08-01
Fix EncryptedServerSideMultipartManagerIT#cantUploadSmallMultipartString

tjcelaya:
Straightforward implementation which fixes the test. Since the nested exception hides the fact that this sort of falls under the realm of "logical errors" I'm proposing we use a new exception type to signal that this is a _particular_ kind of MPU error and retrying the same part indefinitely wouldn't be helpful. Open to suggestions on exception name.


java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
```diff
@@ -440,11 +440,11 @@ public void cantUploadSmallMultipartString() throws IOException {
         for (int i = 1; i < parts.length; i++) {
             final int j = i;
```
dekobon: If j==i, then why is j defined separately here?

tjcelaya: It's used in the lambda expression so it needs to be final.

PR #306, opened 2017-08-15
Documentation Organization

tjcelaya:
The README we currently have is _massive_ and includes links to every kind of example we have in addition to the table of all possible configuration options. I'd like to make the documentation more approachable by breaking it up into a few new files.

Main changes:
 - Reduce list of config options in README.md to the bare minimum required to get started
 - INSTALL.md for requirements, installation, configuration, plus `ConfigContext` subclasses
 - TESTING.md for notes about running and writing tests, and diagnosing skipped tests
 - FAQ.md for misc notes about subusers, MPU, and CSE + MPU
 - Installation links for JCE, libnss, and native FastMD5 from README to relevant section in INSTALL.md

Additionally the sections of the test suite which could be unintentionally skipped have been called out in the `docs/TESTING.md` file

(branch name is no longer relevant)


FAQ.md
```diff
@@ -0,0 +1,82 @@
...
+
+`mvn test` will run unit tests while `mvn verify` will include integration tests and benchmarks.
+
+See [the testing documentation](/TESTING.md) for additional notes about running the test suite.
+
+## How do Accounts/Usernames/Subusers differ?
+
+Joyent's SmartDataCenter account implementation is such that you can have a
+subuser as a dependency upon a user. This is part of SmartDataCenter's [RBAC
+implementation](https://docs.joyent.com/public-cloud/rbac/users). A subuser
```
heyawhite: You don't define users or accounts, you've jumped straight to subusers. Consider breaking it down:

> Joyent's SmartDataCenter account implementation is such that you can have a subuser as a dependency upon a user. This is part of SmartDataCenter's [RBAC implementation](https://docs.joyent.com/public-cloud/rbac/users).
> 
> + **Accounts** - The group created for your company on Triton, managed by a user and subusers.
> + **Users** - Users are login credentials that allow different people in your organization to log in to your Joyent Cloud account. Generally, users are account owners, but this term may be used for account owners and in place of "subusers".
> + **Subusers** - A subuser belongs to the account holder, who maintains control of the subusers. This may be people in your organizations who belong to the larger company account. Subusers have a unique username that is joined with the account holder's username. This is in the format of "user/subuser".

I think the definition for "accounts" definitely could warrant more thought -- I may come back to this. Especially considering "accounts" as thought of for Joyent are not necessarily the same as seen in the Java Manta library.

FAQ.md
```diff
@@ -0,0 +1,82 @@
...
+is a user with a unique username that is joined with the account holder's
```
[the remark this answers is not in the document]
heyawhite: I did use this in the above bulleted list.

FAQ.md
```diff
@@ -0,0 +1,82 @@
...
+username. Typically, this is in the format of "user/subuser".
+
+Within the Java Manta library, we refer to the account name as the entire
+string used to login - "user/subuser". When we use the term user it is in
+reference to the "user" portion of the account name and when we use the term
+subuser, it is in reference to the subuser portion of the account name.
+
+The notable exception is that in the configuration passed into the library,
```
heyawhite: Any chance that could be updated? Probably not, but it would be helpful.

FAQ.md
```diff
@@ -0,0 +1,82 @@
...
+we have continued to use the terminology *Manta user* to refer to the
+account name because of historic compatibility concerns.
+
+## Does Triton Object Storage (Manta) support Multipart Uploads?
```
heyawhite: This is the first time we're seeing that Triton Object Storage is Manta. Note for us, we need to think of a better way to make this transition.

heyawhite: Also, "multipart uploads" is not a proper noun, therefore it should not be capitalized.

FAQ.md
```diff
@@ -0,0 +1,82 @@
...
+   previous step. The order that parts are uploaded does not matter,
+   what does matter is that each part has the appropriate part number specified.
+5. Execute `complete` to commit the parts to the object on the server. At this
+   point the server will assemble the final object.
+
+An example application is provided in the **java-manta-examples** module, named
+[ServerMultipart](/java-manta-examples/src/main/java/ServerMultipart.java),
+to help illustrate the workflow.
+
+## How does Client-side Encryption interact with Multipart Uploads?
```
heyawhite: Same caps critique. Should be:

> How does client-side encryption interact with multipart uploads?

FAQ.md
```diff
@@ -0,0 +1,82 @@
...
+
+* Uploaded parts must be uploaded in sequential order (e.g. part 1,
+  part 2, ...).
+* Different parts can not be uploaded between different JVMs.
+* At this time only the cipher mode AES/CTR is supported.
+
+You will need to configure an instance of MantaClient using the
+settings you would for client-side encryption, namely setting
+`manta.client_encryption` to `true` and configuring the cipher and
+encryption key.  Additionally, you need to create an instance of
```
heyawhite: Is "instance" the best word to use? We use it so often to mean a container of sorts, so if there's a synonym that would be better, let's do it.

FAQ.md
```diff
@@ -0,0 +1,82 @@
+# FAQ
+
+## How do I run the test suite?
+
+`mvn test` will run unit tests while `mvn verify` will include integration tests and benchmarks.
+
+See [the testing documentation](/TESTING.md) for additional notes about running the test suite.
```
heyawhite: Looks like folks should also be viewing the requirements documentation.

FAQ.md
```diff
@@ -0,0 +1,82 @@
+# FAQ
```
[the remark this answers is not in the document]
heyawhite: Assume it's possible someone comes to this page without having read anything else.

Additional possible FAQs:

+ Where do I learn more about Triton Object Storage (Manta)? -- this can be served with a link to the docs
+ What are some other projects using the Java-Manta SDK? -- direct people back to the main README list

Do we have any other notable SDKs and would someone using the Java-Manta SDK care?

TESTING.md
```diff
@@ -0,0 +1,74 @@
+
+# Requirements
+
+In order to run the full test suite users will need access to a running Manta
```
heyawhite: Maybe link to the testing documentation here?

TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+  required: org.apache.http.impl.client.CloseableHttpClient
+  found: com.joyent.manta.org.apache.http.impl.client.CloseableHttpClient
+  reason: argument mismatch; com.joyent.manta.org.apache.http.impl.client.CloseableHttpClient cannot be converted to org.apache.http.impl.client.CloseableHttpClient
+```
+
+The current solution to this issue is to place the test inside
+`java-manta-client/src/integration` and make sure it's included by the
+relevant [testng configuration](/java-manta-client/src/integration/resources/testng-integration.xml).
+
+Surefire and Failsafe plugins are configured to look for TestNG configuration
```
heyawhite: What is TestNG?

TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+ - `manta.key_id` or `MANTA_KEY_ID`
+ - `manta.key_path` or `MANTA_KEY_PATH`
+
+## Multipart Upload (Server-side support)
+Tests specific to server-side MPU functionality will be automatically skipped
+if the server con does not support that capability.
+
+## Roles
+Tests which rely on Triton roles will require the creation of roles through the
+Triton customer portal. The names of the roles used in the tests will be
```
heyawhite: Link out to my.joyent.com or to the docs on how to create roles.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
+# Installation
+
+## Requirements
+* [Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index.html) or higher.
```
cburroughs: I would drop the download url.  Anyone who got this far knows what Java is, and I would prefer to avoid implying that the choice of JDKs matters in general.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+* [Java Cryptography Extension](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html) to
+  use Client-side Encryption with stronger ciphers (192/256-bit).
+* [Maven 3.3.x](https://maven.apache.org/) to contribute to the project.
```
cburroughs: I think we can just say "Maven" at this point.  Also `3.3.x` doesn't match `maven.min.version`

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+
+## From Source
+If you prefer to build from source, you'll also need
+[Maven](https://maven.apache.org/), and then invoke:
+
+``` bash
+$ mvn package
+```
+
+Which will compile the jar to `./targets/java-manta-${VERSION}.jar`. You can then
```
cburroughs: I don't think this is actually correct (artifact name, multi-module).  I think `mvn package` is sufficient since explaining how maven works is not in scope.

TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+following:
+
+```
+  required: org.apache.http.impl.client.CloseableHttpClient
+  found: com.joyent.manta.org.apache.http.impl.client.CloseableHttpClient
+  reason: argument mismatch; com.joyent.manta.org.apache.http.impl.client.CloseableHttpClient cannot be converted to org.apache.http.impl.client.CloseableHttpClient
+```
+
+The current solution to this issue is to place the test inside
+`java-manta-client/src/integration` and make sure it's included by the
```
cburroughs: This isn't true yet, right?

TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Integration Test Path
+Integration tests only touch paths under the integration test root path which
+is configurable using the `manta.it.path` system property or the `MANTA_IT_PATH`
+environment variable. The default behavior is to compute the base path based on
+the user's private folder, e.g. `/my.account/stor/java-manta-integration-tests`.
+
+# Running Tests
+
+Run `mvn verify` from the project root to run all tests. Some Maven goals will
```
cburroughs: I'm not sure what the statement that some goals don't work is referring to.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
+# Installation
+
+## Requirements
+* [Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index.html) or higher.
+* [Java Cryptography Extension](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html) to
+  use Client-side Encryption with stronger ciphers (192/256-bit).
+* [Maven 3.3.x](https://maven.apache.org/) to contribute to the project.
+
+### CLI Requirements
```
cburroughs: I"m not sure what is "CLI" specific here.  "CSE" intended?

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+``` bash
+$ mvn package
+```
+
+Which will compile the jar to `./targets/java-manta-${VERSION}.jar`. You can then
+add it as a dependency to your Java project.
+
+If you want to skip running of the test suite, use the `-DskipTests` property.
+
+# Configuration
```
cburroughs: Naming: I'm a little worried that `INSTALL` is an unexpected place for the configuration details (which is probably the most frequently referenced part of this documentation.)  I know we have several pointers in the README, but as a java developer I suspect I would skim over that because "duh, no need to install, just use maven!".  I'm also not sure there is a all around better or more idiomatic place since this isn't just a simple case of javadoc and method signatures.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+| Default                              | System Property                    | Environment Variable           |
+|--------------------------------------|------------------------------------|--------------------------------|
+| https://us-east.manta.joyent.com:443 | manta.url                          | MANTA_URL                      |
+|                                      | manta.user                         | MANTA_USER                     |
+|                                      | manta.key_id                       | MANTA_KEY_ID                   |
+| $HOME/.ssh/id_rsa                    | manta.key_path                     | MANTA_KEY_PATH                 |
+|                                      | manta.key_content                  | MANTA_KEY_CONTENT              |
+|                                      | manta.password                     | MANTA_PASSWORD                 |
+| 20000                                | manta.timeout                      | MANTA_TIMEOUT                  |
+| 3 (6 for integration tests)          | manta.retries                      | MANTA_HTTP_RETRIES             |
```
cburroughs: Is the "6" part actually true? Why?

README.md
```diff
@@ -191,13 +66,14 @@ you will then construct a `MantaClient` instance. The `MantaClient` class
 is intended to be used per Manta account. It is thread-safe and you should
 share one instance across multiple threads.
 
-For detailed usage instructions consult the provided javadoc and examples.
+For detailed usage instructions consult the provided JavaDoc and examples. The
+JavaDoc can also be browsed online at [javadoc.io](https://javadoc.io/doc/com.joyent.manta/java-manta-client/)
```
cburroughs: I was not aware of this service.  Nice find!

INSTALL.md
heyawhite: I'd argue choice matters is there's something that wouldn't work on an earlier version. Is that the case?

README.md
```diff
@@ -14,171 +14,46 @@ SDK for interacting with Joyent's Manta system.
 
 ## Installation
 
-### Requirements
-* [Java 1.8](http://www.oracle.com/technetwork/java/javase/downloads/index.html) or higher.
-* [Maven 3.3.x](https://maven.apache.org/)
-* [Java Cryptography Extension](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html)
+Java 1.8 is required. The main client module can be installed by adding the following dependency in Maven:
```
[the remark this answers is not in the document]
dekobon: Maven minimum version should be 3.1. It was changed to lower version for compatibility with TravisCI.

INSTALL.md
cburroughs: Sorry, here by "choice" of I meant "OracleJDK" vs "OpenJDK" vs "AzuJDKl" etc, not the version number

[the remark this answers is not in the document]
dekobon: The only JDKs that have been tested are:
* Oracle JDK 1.8
* OpenJDK 1.8
* Azul Zulu 1.8

The JDKs that haven't been tested:
* IBM JDKs
* Oracle, OpenJDK, Azul Zulu 1.9+
* Azul Zing

It is possible that the JDK 1.9 is broken. The last time I tried to test using the OpenJDK 1.9, there were a lot of problems, but I couldn't tell if the problem was on our side or the JDK's.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+flexible combination of parameters from different sources:
+
+  - `StandardConfigContext` leaves all values unset and provides named setters
+  - `EnvVarConfigContext` reads values from environment variables (e.g. `export MANTA_URL=value`)
+  - `SystemSettingsConfigContext` reads values from system properties (e.g. from flags `-Dmanta.url=value` or a `.properties` file)
+  - `ChainedConfigContext` allows the combination of multiple `ConfigContext`s
+
+## Parameters
+
+| Default                              | System Property                    | Environment Variable           |
```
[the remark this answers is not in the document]
dekobon: Why would you put the default in the middle between two settable properties?

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+- `manta.client_encryption` - set to `true`
+- `manta.encryption_key_bytes_base64` or `manta.encryption_key_bytes`
+
+Additionally, you should set the following system properties to support encrypting and uploading files using client-side
+encryption.
+
+- `manta.encryption_algorithm`
+- `manta.encryption_key_id`
+
+Below is a table of each of the supported encryption algorithms and the features they provide.
```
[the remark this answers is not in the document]
dekobon: That's a crazy big question. How would you address this question:
> Where can I learn more about particle physics?

I mean this in all seriousness. Should we link out to an simple explanation page on the internet? This might give folks a false sense of confidence.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+| AES192/CTR/NoPadding                 | Yes                                | No                             |
+| AES256/CTR/NoPadding                 | Yes                                | No                             |
+|                                      |                                    |                                |
+| AES128/CBC/PKCS5Padding              | No                                 | No                             |
+| AES192/CBC/PKCS5Padding              | No                                 | No                             |
+| AES256/CBC/PKCS5Padding              | No                                 | No                             |
+
+*Note* that all algorithms will honor the `manta.encryption_auth_mode` setting.
+
+*Note* that any non-authenticated encryption algorithm will have an HMAC generated to
```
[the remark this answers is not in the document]
dekobon: A link to Wikipedia may be fine.

INSTALL.md
```diff
@@ -0,0 +1,315 @@
...
+authenticate that ciphertext was not modified.
+
+*Note* each instance of a MantaClient will only support the encryption/decryption for the algorithm configured for that instance.
+
+### Improving Encryption Performance
+
+#### Enabling libnss Support via PKCS11
+
+[Libnss](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS) can be used with
```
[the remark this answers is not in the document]
dekobon: That is a mistake.

FAQ.md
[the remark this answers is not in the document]
dekobon: Almost all of the other SDKs use an environment variable called MANTA_SUBUSER. I was told at the time of writing support that this way of defining a subuser was being deprecated. By the time I had finished, that decision had been reversed.

FAQ.md
dekobon: Unfortunately, the word instance in this case has more common usage than referring to a "compute instance". The term instance has a very precise meaning in OOP it is referring to an instantiated class. See: https://alfredjava.wordpress.com/2008/07/08/class-vs-object-vs-instance/

INSTALL.md
dekobon: This should be Maven 3.1

INSTALL.md
dekobon: CLI is referring to the Java module java-manta-cli. When you are running it stand-alone it has different runtime requirements.

INSTALL.md
dekobon: Maybe we should have a doc called USAGE?

INSTALL.md
dekobon: It was true. I don't know if it had changed. It was high to deal with the high failure rate when testing against JPC Manta.

TESTING.md
dekobon: I don't remember. We need to reverify, but I do remember some goals don't work when manually executed.

INSTALL.md
dekobon: Also, yes - it won't work on any version earlier than 1.8.

INSTALL.md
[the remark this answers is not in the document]
heyawhite: Good point. What about:

```
| System Property | Environment Variable | Default | 
```

I standby System Property being first -- it better fits the standards of how people read tables, with the items being defined across the top and down on the left.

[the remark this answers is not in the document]
dekobon: The only reason I did default first was because there is a weird truncation issue on many people's browsers when the view the table on github and they have to scroll right. I thought that the default value was the most important piece of information in that table and it shouldn't ever be hidden.

I know that is a strange thought process, but that's what it was. Honestly, I don't care that much. Let's do it at the end.

INSTALL.md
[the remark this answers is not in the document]
heyawhite: I wouldn't say that it requires a simple explanation link-out. It is a really, really big question. Personally, I want to learn more and I think I would better be able to phrase that question if I could better articulate not just what they are but how they work.

I would assume that most people who have gotten to this point know what they need to know about encryption, but I want to know what you all find to be the most trustworthy resources on the topic. Yes, I can read the wikipedia page, but beyond that, I want to know why one may be a better choice than another. What resources can you point me to?

INSTALL.md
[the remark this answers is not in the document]
heyawhite: Fine by me, it's definitely not a priority.

README.md
```diff
@@ -206,260 +82,31 @@ using the `MantaJobBuilder` class. `MantaJobBuilder` provides a fluent interface
...
-name = NSS
-nssLibraryDirectory = /usr/lib/x86_64-linux-gnu
-nssDbMode = noDb
-attributes = compatibility
-```
-
-Make sure that the name field is `NSS` because the SDK will only use the library if that specific
-name is set. Next, edit the following file: `$JAVA_HOME/jre/lib/security/java.security`
+* [Jobs using MantaClient](/java-manta-examples/src/main/java/JobsWithMantaClient.java)
+* [Jobs using MantaJobBuilder](/java-manta-examples/src/main/java/JobsWithMantaJobBuilder.java)
```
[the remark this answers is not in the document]
tjcelaya: Just to clarify, you're asking for examples of creating jobs using MantaClient or MantaJobBuilder? Not sure how the portal relates to these code samples.

TESTING.md
tjcelaya: Please clarify, this is the testing documentation file.

TESTING.md
tjcelaya: `mvn validate` for example:

`Could not resolve following dependencies: [com.joyent.manta:java-manta-benchmark:jar:3.1.7-SNAPSHOT (test)]`

INSTALL.md
tjcelaya: Name was inspired by autoconf top-level names. Switching to USAGE

INSTALL.md
tjcelaya: I can't find anything in code that would set the higher default. I'll verify the actual number by running a test case and remove the parenthetical if it's no longer different.

TESTING.md
tjcelaya: Not yet no, I have a branch that adds just a single test which works this way that I need to merge before this is accurate.

README.md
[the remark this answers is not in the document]
heyawhite: It may not -- Manta is not my expertise and I made an assumption that the two were related. I'll take this as a "nope."

TESTING.md
heyawhite: Whoops, I thought I made that comment on another doc. Scratch this.

TESTING.md
tjcelaya: The testing framework we use, I linked Surefire (the unit test runner), Failsafe (integration test runner) and TestNG (library for defining tests) to their relevant docs.

TESTING.md
tjcelaya: Retracted the paragraph in favor of "we're going to provide a workaround soon" and I'll update the text with that PR.

tjcelaya:
Incorporated feedback and responded to questions. Merging later today unless there are any complaints.

PR #308, opened 2017-08-16
use an option parsing library for java-manta-cli

cburroughs:
java-manta-cli had atrophied and the documented help behavior no longer
matched reality.  Sub-commands such as validate-key could no longer be
run as the arguments checked did not match the arguments used.  In
practice the commands were also difficult to use due to the copious
log messages.

Using a library instead of parsing args by hand allows all this to be
cleaned up and paves the way for smoothly adding more commands.

### DIFF
--- a/java-manta-cli/pom.xml
+++ b/java-manta-cli/pom.xml
@@ -39,6 +39,11 @@
             <artifactId>commons-lang3</artifactId>
             <version>${dependency.commons-lang.version}</version>
         </dependency>
+        <dependency>
+            <groupId>info.picocli</groupId>
+            <artifactId>picocli</artifactId>
+            <version>${dependency.picocli.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
+++ b/java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
@@ -7,6 +7,8 @@
  */
 package com.joyent.manta.cli;
 
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
 import com.joyent.manta.client.MantaClient;
 import com.joyent.manta.client.MantaObject;
 import com.joyent.manta.client.MantaObjectResponse;
@@ -17,9 +19,8 @@ import com.joyent.manta.config.DefaultsConfigContext;
 import com.joyent.manta.config.EnvVarConfigContext;
 import com.joyent.manta.config.MapConfigContext;
 import com.joyent.manta.util.MantaVersion;
-import org.apache.commons.lang3.ArrayUtils;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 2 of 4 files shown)


cburroughs:
Examples:
````
$ java -jar target/java-manta-cli-3.1.6-SNAPSHOT-jar-with-dependencies.jar -h
                .     .             
                |_.-._|             
              ./       \.          
         _.-'`           `'-._      
      .-'        Java         '-.   
    ,'_.._       Manta       _.._', 
    '`    `'-.           .-'`    `' 
              '.       .'           
                \_/|\_/           
                   |                
                   |                
                   |                

Usage: java-manta-cli [-hv]

Basic manta cli commands using the java-client. Not a stable interface.

Options:
  -v, --version
  -h, --help
Commands:
  connect-test  Try to connect
  dump-config   Dumps the configuration used for configuring a Manta client.
  generate-key  Generate an encryption key
  ls            list
  get-file      Performs a download of file in Manta
  put-file      Performs a put of a local file to Manta
  validate-key  Validate an encryption key
```

```
$ java -jar target/java-manta-cli-3.1.6-SNAPSHOT-jar-with-dependencies.jar connect-test --help
Usage:

Try to connect

connect-test [-h] [--log-level=<logLevel>]

Description:

Attempts to connect to Manta using system properties and environment variables
for configuration

Options:
      --log-level=<logLevel>  TRACE, DEBUG, INFO, WARN(default), ERROR
  -h, --help
```

java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
```diff
@@ -17,364 +19,367 @@
...
 import javax.crypto.SecretKey;
 import javax.crypto.spec.SecretKeySpec;
 
-/**
- * Class providing a CLI interface to the Java Manta SDK.
- *
- * @author <a href="https://github.com/dekobon">Elijah Zupancic</a>
- * @since 3.0.0
- */
+@CommandLine.Command(name = "java-manta-cli", sortOptions = false,
```
dekobon: Does picocli drop colors when in non-interactive mode?

java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
```diff
@@ -17,364 +19,367 @@
...
-        b.append(INDENT).append(ConfigContext.toString(config)).append(BR);
+            b.append("Creating new connection object").append(BR);
+            try (MantaClient client = new MantaClient(config)) {
+                b.append(INDENT).append(client).append(BR);
 
-        b.append("Creating new connection object").append(BR);
-        try (MantaClient client = new MantaClient(config)) {
-            b.append(INDENT).append(client).append(BR);
+                b.append("Attempting GET request to: ").append(filePath).append(BR);
+                b.append("\nMetadata values: \n").append(client.get(filePath).getMetadata().toString());
```
dekobon: You may want to swap \n with BR.

java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
cburroughs: "Picocli will only emit ANSI escape codes on supported platforms."

If I pipe to a file all  looks well (no-colors or malformed escape codes).

java-manta-cli/pom.xml
```diff
@@ -17,6 +17,10 @@
     <artifactId>java-manta-cli</artifactId>
     <packaging>jar</packaging>
 
+    <properties>
+        <dependency.picocli.version>0.9.8</dependency.picocli.version>
```
tjcelaya: Even if it's only relevant to the cli module I'd prefer we keep all version properties in the parent pom. I noticed the benchmark module does this for JMH as well and am not asking you to clean those up, just to move this one to the parent pom if possible.

java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
```diff
@@ -17,364 +19,367 @@
...
-    /**
-     * Entrance to the CLI.
-     * @param argv parameters passed.
-     */
-    public static void main(final String[] argv) {
-        if (ArrayUtils.isEmpty(argv) || StringUtils.isBlank(argv[0])) {
-            System.out.println(help());
+    public static void main(final String[] args) {
+        final CommandLine application = new CommandLine(new MantaCLI());
+        application.registerConverter(Path.class, s -> Paths.get(s));
```
tjcelaya: nit: this could be written as

`application.registerConverter(Path.class, Paths::get);`

java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
```diff
@@ -17,364 +19,367 @@
...
+            }
+            System.out.println(b.toString());
+        }
     }
 
+
     /**
      * Dumps the configuration used for configuring a Manta client.
      * @return String containing the contents of the parsed configuration
      */
```
tjcelaya: Did you intend to keep this method JavaDoc? They seem to have been removed from the other commands once they became classes.

cburroughs:
Landing with suggested changes taken up.

Issue #247, opened 2017-05-25
Certain ConfigContexts cause NullPointerExceptions despite having documented defaults


tjcelaya:
**TL;DR: we aren't handling some configuration parameters in the execution path which have documented default values**

Despite the defaults listed in [README.md](README.md#configuration), certain implementations of `ConfigContext` result in `NullPointerException`s. Adding getters for `disableNativeSignatures` and `noAuth` to these classes would save some trouble for developers that choose to use these classes. One option is to implement getters in these classes if the behavior of null is best decided by the class itself. Most of these classes use `MantaUtils.parseBooleanOrNull`

~Problematic Classes~ Classes that do what they say they do:
 - `MapConfigContext`: uses `MantaUtils.parseBooleanOrNull` which returns null when passed null
 - `EnvVarConfigContext`: uses `MantaUtils.parseBooleanOrNull` as well
 - `StandardConfigContext`: doesn't implement anything, which is probably as intended and implies the null-handling for default behavior belongs outside these classes
 - ~`ChainedConfigContext`~: abstracts over arrays of multiple `ConfigContext`s and probably shouldn't be concerned with this kind of thing

Alternatively, we could modify the lines that cause the `NullPointerException`s codepath to check handle nulls explicitly, specifically the following two lines
 - https://github.com/joyent/java-manta/blob/master/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java#L207
 - https://github.com/joyent/java-manta/blob/master/java-manta-client/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java#L171 Interestingly enough line 165 in that same file does explicit null checking for `useNativeCodeToSign`

Code used to verify which classes throw and which don't:

https://gist.github.com/tjcelaya/df63bfe932afff07de4f501d6daff241

dekobon:
The fundamental problem is that we don't have a stable value for **UNSET**. I could see us refactor to a wrapping object that would allow for that information to be stored, but that would be a fair bit of work.

PR #309, opened 2017-08-16
Update the config context for the SimpleClient example

kellymclaughlin:
I was exploring the java client and I realized that the `SimpleClient` example does
not work as it currently stands. I updated it to use a `ChainedConfigContext` similar to other
examples.

### DIFF
--- a/java-manta-examples/src/main/java/SimpleClient.java
+++ b/java-manta-examples/src/main/java/SimpleClient.java
@@ -7,8 +7,9 @@
  */
 
 import com.joyent.manta.client.MantaClient;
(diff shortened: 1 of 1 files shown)


tjcelaya:
The missing default value for `disableNativeSignatures` is documented in #247 (which has been neglected) but it doesn't hurt to fix the example in the meantime.

I'd prefer this change without the wildcard import but it's probably fine in an example.

kellymclaughlin:
I can definitely make the imports explicit. I'll push that change.

tjcelaya:
Thanks @kellymclaughlin

PR #313, opened 2017-08-17
Resolve #247

### LINKED ISSUE #247: Certain ConfigContexts cause NullPointerExceptions despite having documented defaults

tjcelaya:
Handle potentially-NPE-inducing configuration params that have documented defaults

See #247

### DIFF
--- a/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -39,6 +39,7 @@ import com.joyent.manta.util.MantaUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 import org.apache.http.HttpEntity;
@@ -205,7 +206,7 @@ public class MantaClient implements AutoCloseable {
         final KeyPair keyPair = keyPairFactory.createKeyPair();
 
         final Signer.Builder builder = new Signer.Builder(keyPair);
-        if (config.disableNativeSignatures()) {
+        if (ObjectUtils.firstNonNull(config.disableNativeSignatures(), false)) {
             builder.providerCode("stdlib");
         }
         final ThreadLocalSigner signer = new ThreadLocalSigner(builder);
--- a/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -14,6 +14,7 @@ import com.joyent.manta.client.crypto.ExternalSecurityProviderLoader;
 import com.joyent.manta.client.jobs.MantaJob;
 import com.joyent.manta.client.jobs.MantaJobBuilder;
 import com.joyent.manta.client.jobs.MantaJobError;
+import com.joyent.manta.config.BaseChainedConfigContext;
 import com.joyent.manta.config.ConfigContext;
 import com.joyent.manta.config.DefaultsConfigContext;
 import com.joyent.manta.config.KeyPairFactory;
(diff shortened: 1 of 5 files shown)


java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -205,7 +206,7 @@ public MantaClient(final ConfigContext config) {
         final KeyPair keyPair = keyPairFactory.createKeyPair();
 
         final Signer.Builder builder = new Signer.Builder(keyPair);
-        if (config.disableNativeSignatures()) {
+        if (ObjectUtils.firstNonNull(config.disableNativeSignatures(), false)) {
```
dekobon: I suggest that you add a constant field to `DefaultsConfigContext`, so that you reference the default value rather than explicitly set it here.

java-manta-client/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
```diff
@@ -160,15 +160,10 @@ public MantaConnectionFactory(final ConfigContext config,
...
-        final boolean useNativeCodeToSign;
-
-        useNativeCodeToSign = config.disableNativeSignatures() == null
-                || !config.disableNativeSignatures();
-
         final HttpSignatureAuthScheme authScheme;
 
         // If we have auth disabled, then we don't assign any signer classes
-        if (config.noAuth()) {
+        if (ObjectUtils.firstNonNull(config.noAuth(), false)) {
```
dekobon: Same comment as above.

java-manta-client/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
```diff
@@ -312,7 +312,7 @@ public String getEncryptionPrivateKeyPath() {
     public void overwriteWithContext(final ConfigContext context) {
         /* If a default context is being used to overwrite after this
          * context has been initialized, then we want to be careful to not
-         * overwrite values that have already been set with defaults. */
+         * overwrite values that have already been set by non-default contexts. */
         boolean isDefaultContext = context instanceof DefaultsConfigContext;
```
dekobon: Do we want this logic to apply for all subclasses of DefaultConfigContext? I don't know what would be right or wrong, but that would be the inflection point to decide between .equals() and instanceof.

tjcelaya: While it's unlikely that someone would extend `DefaultsConfigContext` we could make that class `final`. We'd lose the ability to mock that class using plain Mockito but it would enforce the expectation that defaults are only coming from one place.

dekobon: Or you could just check equality with .class.equals()

tjcelaya:
@dekobon Created constants instead of using `BaseConfigContext.DEFAULT_CONFIG.foo()`

Although I did switch to `.getClass().equals()` for the statement you were asking about I want to leave `DefaultsConfigContext` as `final` since I don't think we should allow consumers to extend it.

dekobon:
I almost always error on the side of never declaring classes final because as an integrator, I've had to do so much black magic in the past because the author of a library couldn't foresee that someone might want to extend a class.

tjcelaya:
I guess I could envision a really weird situation where someone might need to inject configuration through the `DefaultsConfigContext` class, changed back.

Issue #317, opened 2017-08-24
code coverage for integration tests; make coverage more visible


cburroughs:
We currently generate code coverage information for unit tests but only generate a pretty report when the `site` goal is run. (So rarely, and not during the normal development workflow).

It would be better to:
 * (A) Always produce the report when tests are run (ie as part of the `test` goal).  This appears straightforward.
 * (B) Generate reports for integration tests.  This has not been straightforward.

To generate a report, jacoco needs access to the class files and reasonably expects them to be in `java-manta-client/target/classes`.  However, because of (surprise!) our shade+relocation step the modified class files are only in the jar file.  How jacoco keeps tracks of classes has its own page <http://www.jacoco.org/jacoco/trunk/doc/classids.html> and the interaction with other things that muck with bytecode is unhappy.

Options moving forward:
 1.  Keep trying to make it work through maven.  I can make a report manually (using the new jacoco cli), so it seems like it must be possible to do this somehow.  But I don't have a good guess as to why what I have tried is any different from the cli. (I have also verified that things "Just Work" with shading disabled.)
 2. Give up for now and put some instructions in the readme for how to manually generate the integration test reports.  I think in practice no one would look at this with any regularity.
 3. Move all integration tests to the main module.  I think we are been burned by oddities of the shade+relocation often enough that there is real value in testing the same thing we are shipping.
 4. Some type of profile/conditional to make shading optional; CI runs integration tests twice.  This might also give us an escape valve for #307

cburroughs:
Some progress (internal):
 * https://jenkins.joyent.us/job/java-manta-it-coverage-demo/ws/java-manta-client/target/site/jacoco/index.html
 * https://jenkins.joyent.us/job/java-manta-it-coverage-demo/ws/java-manta-it/target/site/jacoco-it/index.html

Still missing source code linking for the integration tests, and the reports are not that useful without them.  At first look, do the integration reports appear *correct*? I would have expected our integration coverage percent to by higher than our unit percent.

cburroughs:
After further hanging on option 1 above I have something that works.  The above links should now have working code coverage reports.  The workarounds are not super pretty but I think any problems would be confined to java-manta-it and not downstream users.

>  I would have expected our integration coverage percent to by higher than our unit percent.

I'm not sure what I was looking at before, but the integration test coverage is showing as higher, matching my intuition.

cburroughs:
Example runs (internal) after landing:
[...]

Issue #307, opened 2017-08-16
Provide artifacts with no shaded dependencies


dkocher:

cburroughs:
Could you elaborate on your use case for the alternative artifact?  Note that the shaded classes are also relocated, so there should be no conflict between (for example) `org.apache.http.client.HttpClient` and `com.joyent.manta.org.apache.http.client.HttpClient`.

dkocher:
We don't want to have these libraries duplicated but manage versions of transitive dependencies ourselves.

dekobon:
I see no technical reason why this isn't possible. The only gotcha is the dependency `com.twmacinta:fast-md5` that is coming from the `ci-jenkins` repository and not the oss-sonatype repository. We may need some Maven magic to properly embed the dependency all the while providing all of the other dependencies without shading.

The rational for shading has been that we haven't wanted to spend time on support when users have convergent dependency problems. Working on Java projects over the years, I have burned countless hours from user bug reports because of dependency convergence and users who are inexperienced with Maven.

My original intent with shading Apache Http Client was simply to remove the entire problem because most users can function just fine with a shaded artifact and it reduced our matrix of pain.

tjcelaya:
Just to outline what our options are:

1. Bump the major version and publish `java-manta-client` as the unshaded artifact. This might be an unexpected change for users going from 3.1.6 to 4.0.0 and suddenly having dependency conflicts or pulling in previously-shaded dependencies.
2. Move the actual code in `java-manta-client` into an "internal" module and continue publishing the shaded artifact as `java-manta-client`. I was able to find a [few projects](https://search.maven.org/#search%7Cga%7C1%7Cunshaded) which publish a separate "$ARTIFACT-unshaded" artifact but it seems highly irregular. This might also clobber change attribution I believe.
3. Create an alternate profile that the `java-manta-client` module can use to change its artifact name and disable shading. This seems cumbersome and prone to breakage, e.g. do maintainers need to perform the release process twice, once for the `release` profile and again for the custom profile?

Regarding `com.twmacinta:fast-md5` coming from the `ci-jenkins` repository. Does that mean we _always_ need shading in the main module just for that particular dependency, or would say, `maven-assembly-plugin` be a better fit?

dekobon:
[...]

PR #326, opened 2017-08-29
always produce coverage reports for unit & integration tests

cburroughs:
Previously code coverage reports were only generated as part of the
`site` goal, and were thus rarely looked at.  Binding the reports to
the same phase ensures they are always available for inspection.

Due to shading, several hoops need to be jumped through so that the
integration test report can refer to the source and classes under test.
Actions are bound to several maven phases to move files around, acting
as if java-manta-client was part of the integration test module.

ref #317

### DIFF
--- a/java-manta-it/pom.xml
+++ b/java-manta-it/pom.xml
@@ -112,7 +112,180 @@
     </dependencies>
 
     <build>
+        <!-- To generate code coverage reports (of either unit or
+             integration tests) jacoco needs access to both the
+             original (pre-instrumentation) class files, and the
+             source code.
+             <http://www.jacoco.org/jacoco/trunk/doc/classids.html>
+             describes how jacoco keeps track of classes in further
+             detail.  There exist goals that can traverse a
+             multi-module project and aggregate this information from
+             the standard maven locations, but they are confounded by
+             the shading+relocation step.  The final classes shipped
+             in the client jar are *only* in the shaded client jar,
+             and not in target/classes.
+
+             To work around this and provide the necessary information
+             requires jumping through several hoops to act as if the
+             java-manta-client source files were part of the
+             integration test module:
+
+              * Add a new source location to java-manta-it to hold the
+              "upstream" .java files from java-manta-client
+              * Copy the source files there
+              * Make sure the source are removed when not needed so
+              they are never actually compiled.
+              * Unpack the final relocated class files so they are
+              available outside of just a jar.
+
+             This scheme depends on there being no "src/main" classes
+             in this integration test module.
+        -->
         <plugins>
+            <plugin>
+                <!-- This property is generated so we can refer to
+                     another module in a multi-module project without
+                     brittle relative paths. -->
+                <groupId>org.commonjava.maven.plugins</groupId>
+                <artifactId>directory-maven-plugin</artifactId>
+                <version>${maven-directory-maven-plugin.version}</version>
+                <executions>
+                    <execution>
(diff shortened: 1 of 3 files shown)


dekobon:
Would it be simpler to do this change after #307?

java-manta-client/pom.xml
```diff
@@ -234,31 +234,31 @@
                     <relocations>
                         <relocation>
                             <pattern>com.joyent.http</pattern>
-                            <shadedPattern>com.joyent.manta.http</shadedPattern>
```
dekobon: What's the rational to change the shaded pattern? I'll admit the original package name was suboptimal.

If you want to change it, you will need to look for the code points where we detect if we are using a shaded artifact or not. If I recall correctly, this is being done in the kryo module and it may be done elsewhere. In those places, there is a string that refers to the shaded package name.

cburroughs:
* If #307 means we don't shade *at all* then everything complicated here goes away.  But given current use & testing by customers with shaded, I'm not sure that would be a wise path.
 * If #307 means we provide and support both options equally then this probably gets even more complicated.  Would we have two integration test modules that shared test cases? I'm not sure what that would look like.
 * If #307 means there is a `-noshade` jar we provide as an on-your-own escape hatch, I don't think it has any effect on this change.

dekobon:
If you change the shading pattern, you will need to change this line: https://github.com/joyent/java-manta/blob/master/java-manta-client-kryo-serialization/src/main/java/com/joyent/manta/serialization/EncryptionStateSerializer.java#L110

cburroughs:
Kept the existing shaded paths, and adding corresponding exclusions.

cburroughs:
This is rebased after the Great Unshading and is ready to look at again:

Examples (internal):
 * https://jenkins.joyent.us/job/java-manta-it-coverage-demo/ws/java-manta-client/target/site/jacoco/index.html
 * https://jenkins.joyent.us/job/java-manta-it-coverage-demo/ws/java-manta-it/target/site/jacoco-it/index.html

java-manta-it/pom.xml
```diff
@@ -112,8 +112,181 @@
...
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>com.joyent.manta</groupId>
+                                    <artifactId>java-manta-client</artifactId>
+                                    <type>jar</type>
+                                    <overWrite>true</overWrite>
+                                    <outputDirectory>${project.build.directory}/classes</outputDirectory>
+                                    <includes>**/*.class</includes>
+                                    <excludes>com/joyent/manta/com/**,com/joyent/manta/org/**,com/joyent/manta/io/**,com/joyent/manta/http/signature/**</excludes>
```
tjcelaya: Just out of curiosity, under what circumstances would this `<excludes>` list change? They seem to be the relocation roots but I just want to be sure I know how to maintain this. I'm assuming coverage for library classes shows up if this isn't set?

cburroughs: Yep, this is just working backwards from the relocation roots in `java-manta-client/pom.xml`.  Additions there may also require changes here.  At some point (4.0?) I think it would be worthwhile to have only a single package path for relocated (ex `com.joyent.manta.shade`) just to keep things a bit tidier.

tjcelaya: Agreed about the single package relocation. The question of "why isn't everything shaded to the same place?" has crossed my mind once before. I'll file an issue for this.

Issue #324, opened 2017-08-29
Excessive Logging During Tests


tjcelaya:
There is a flood of log output that occurs when running the test suite which provides very little useful information. Thousands of lines like the following which don't provide any context and make it more difficult to find actual test failures:

- `[main] DEBUG com.joyent.manta.client.crypto.EncryptingEntityHelper - Creating new OutputStream for multipart [class org.bouncycastle.jcajce.io.CipherOutputStream]`
- `[main] DEBUG com.joyent.manta.client.crypto.EncryptingEntity - HMAC: 699884b09e0fe8f6f467a29cf49f89b7`
- `Exception thrown: Unable to authenticate object's ciphertext. It failed the authenticity checksum. Check to see that the object's binary data hasn't been modified.` (this is intentional, but it looks like a test failure if you're scanning the output)

I'd like to propose the following options:

- raise the default logging level to `INFO` or `WARN` during testing
- change these logging statements to a higher log level if they are still desired
- remove the logging where it doesn't actually provide any value, e.g. `Creating new OutputStream`

cburroughs:
Thanks for bringing this up.  My inclination is to leave the testing log level at `DEBUG`.  If a test fails (especially in a transient fashion!) I'd err on the side of retaining information to figure out why.

For these offending lines, I think a combination of removing them or setting them to `TRACE` is fine.  As you mentioned it looks like a few lines are responsible for the vast majority of output.

cburroughs:
Is this complete with #327 in master?

tjcelaya:
Yep I just forgot to close this out yesterday, thanks for the reminder!

PR #327, opened 2017-08-29
Provide Unshaded Artifact

tjcelaya:
See #307 

So far only the barebones changes have been made, still need to figure out how we're handling `com.twmacinta:fast-md5`.


tjcelaya:
It doesn't look like this will work, the shaded dependencies are missing from the dependency tree (this also results in a runtime error `java.lang.NoClassDefFoundError: io/mikael/urlbuilder/util/Encoder`)

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ j ---
[INFO] co.tjcelaya.sandbox:j:jar:1.0-SNAPSHOT
[INFO] +- com.joyent.manta:java-manta-client:jar:unshaded:3.1.7-SNAPSHOT:compile
[INFO] |  +- org.bouncycastle:bcpkix-jdk15on:jar:1.56:compile
[INFO] |  +- org.bouncycastle:bcprov-jdk15on:jar:1.56:compile
[INFO] |  +- com.squareup.jnagmp:bouncycastle-rsa:jar:2.0.0:compile
[INFO] |  |  \- com.squareup.jnagmp:jnagmp:jar:2.0.0:compile
[INFO] |  |     \- net.java.dev.jna:jna:jar:4.0.0:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.21:compile
[INFO] +- org.slf4j:slf4j-simple:jar:1.7.25:compile
[INFO] +- commons-cli:commons-cli:jar:1.4:compile
[INFO] +- commons-io:commons-io:jar:2.5:compile
[INFO] \- org.apache.commons:commons-lang3:jar:3.4:compile
```

(The `org.slf4j:slf4j-simple`, `commons-cli`, `commons-io`, `commons-lang3` dependencies are from my project)

java-manta-client-unshaded/pom.xml
```diff
@@ -0,0 +1,284 @@
...
+        </dependency>
+
+        <!-- High speed MD5 implementation -->
+        <dependency>
+            <groupId>com.twmacinta</groupId>
+            <artifactId>fast-md5</artifactId>
+            <version>${dependency.fast-md5.version}</version>
+        </dependency>
+
+        <!-- Logging bridge that will be shaded to bridge between the Apache
```
tjcelaya: @dekobon  I'm still unclear on how commons-logging, slf4j, and logback interact with shading. How do we handle this logging bridge library (jcl-over-slf4j) now that shading is in a separate module from the main codebase?

Would I be able to verify that this is working correctly by setting up a logback configuration in a separate project and making sure httpclient logs are visible or is there something in code I can run?

dekobon: slf4j is not shaded. jcl-over-slf4j is shaded because it now ingests logs from shaded sources (all of the commons logging APIs are now shaded).

As for verification, yes I believe what you are suggesting will work - setting up an example project.

tjcelaya:
The current state of this branch does not seem to work with `jcl-over-slf4j` but the downstream project _does_ see the bouncycastle dependency correctly. Still need to fiddle with this some more but I'd appreciate help with the that logging dependency.

cburroughs:
What steps should I do to reproduce the `jcl-over-slf4j` problem?

tjcelaya:
@cburroughs I just wasn't configuring logback correctly. Switching to a SNAPSHOT build from master helped me figure out what was actually wrong: I was using the wrong package name in my `<logger>`

tjcelaya:
Unfortunately with these changes the `java-manta-client` artifact can't use `<minimizeJar>` so the artifact size will increase from 3.2MB to 4.7MB. This is because `<minimizeJar>` looks at the class sources for the project being shaded and omits unreferenced classes. Any suggestions for how to reduce the impact of having to manually `<include>` artifacts during shading would be appreciated.

Some tangentially-related changes that were also made:
- lowering `debug` statements to `trace` and removing some redundant `println`s as discussed in #324. Scanning the maven build output for warnings and plugin activity became much easier once the test suite stopped flooding my console. I can move these into a separate PR if desired.
- Without `<classpathDependencyExcludes>` in the integration test module it was possible to run the test suite **even though the shaded `java-manta-client` artifact was empty**. Discovered that Failsafe will add all transitive dependencies to the test runner classpath, so `java-manta-it` could still see `java-manta-client-unshaded`.

@cburroughs @dekobon Ready for review

cburroughs:
* The kyro module has specific reference to the shaded class names, so I'm not sure that having it depend on `java-manta-client-unshaded` will "just work".
 * Should the kyro module have a `provided` dep so it can't bring the wrong client along?
 * I'm unclear on `classpathDependencyExcludes` in the integration-tests.  Isn't the unshaded jar already excluded from by the dependency reduced pom? This is from failsafe doing additional mucking with the CLASSPATH?
 * I'm puzzled by the ` <skip>true</skip>` for unit tests.  If no tests in a module caused a failure, why is the same thing not needed in `-cli` or `-examples`?

dekobon:
* Can you link out to the specific dependencies on shaded classes?
* Yes, it may be a good idea to specify provided.
* I believe that we were using `classpathDependencyExcludes` in order enforce that the unshaded jar isn't brought in (in particular with IDEs), but you will want to verify this.
* I believe that `<skip>true</skip>` in the integration test module was done to ensure that only the failsafe goal was run and that nothing should ever be run via surefire.

tjcelaya:
The kyro module has specific reference to the shaded class names, so I'm not sure that having it depend on java-manta-client-unshaded will "just work".

- Can you link out to the specific dependencies on shaded classes?

  - [findCloseShieldStreamClass](https://github.com/tjcelaya/java-manta/blob/master/java-manta-client-kryo-serialization/src/main/java/com/joyent/manta/serialization/EncryptionStateSerializer.java#L102) checks both the shaded and unshaded class names. Exercising the kryo module from a separate project works since most of the classes that  module interacts which are not relocated. Could you give an example of a class you would expect to fail or behave incorrectly?

Should the kyro module have a provided dep so it can't bring the wrong client along?

- Yes, it may be a good idea to specify provided.

  - To be clear, the suggestion is to add a `provided` dependency on `java-manta-client` from the kryo module? I can add that just fine but I'm curious how the kryo module could "bring the wrong client along." I don't see any specific need for the kryo module to be aware (on the pom level) that the shaded module exists at all.

I'm unclear on classpathDependencyExcludes in the integration-tests. Isn't the unshaded jar already excluded from by the dependency reduced pom? This is from failsafe doing additional mucking with the CLASSPATH?

- I believe that we were using classpathDependencyExcludes in order enforce that the unshaded jar isn't brought in (in particular with IDEs), but you will want to verify this.

  - During the process of splitting this up I was actually encountering  errors caused by duplicate class definitions. By inspecting failsafe-generated META-INF I discovered that transitive child dependencies are added to the test runner's classpath, so `java-manta-it` had direct access to `java-manta-client-unshaded`

I'm puzzled by the <skip>true</skip> for unit tests. If no tests in a module caused a failure, why is the same thing not needed in -cli or -examples?
    
- I believe that <skip>true</skip> in the integration test module was done to ensure that only the failsafe goal was run and that nothing should ever be run via surefire.

  - `<skip>true</skip>` was added to `java-manta-client` because the maven build would inexplicably fail with a `NoClassDefFoundError` when  invoking the `test` stage (for a class in the unshaded package). I can't seem to reproduce this anymore so I've reverted that change, apologies  for the confusion.

java-manta-client-unshaded/pom.xml
```diff
@@ -0,0 +1,284 @@
...
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>${dependency.slfj.version}</version>
+            <exclusions>
+                <exclusion>
+                    <artifactId>org.slf4j</artifactId>
+                    <groupId>slf4j-api</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
```
dekobon: In the unshaded artifact, you don't want to include this dependency. Compliance with the slf4j APIs shouldn't be enforced for consumers of the library if you are providing an unshaded artifact.

java-manta-client-unshaded/pom.xml
```diff
@@ -0,0 +1,284 @@
...
+        <!-- Library for making HTTP connections -->
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>${dependency.apache-http-client.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
```
dekobon: In the unshaded artifact, you don't want to exclude commons-logging. However, you will want to do this in the shaded module and bring in the jcl-over-slf4j instead of the common-logging dependency.

tjcelaya:
Moved the `commons-logging` exclusion and the `jcl-over-slf4j` dependency to `java-manta-client` in addition to calling out the unshaded package in the readme and the logging requirements.

Confirmed in a downstream project that logs for Apache HttpClient will show up if the user adds a dependency on `jcl-over-slf4j`.

Issue #312, opened 2017-08-17
cli get-file hard to use for large/binary files


cburroughs:
Because java-manta remains the only mainline implementation of client-side-encryption, I have ended up using the cli get/put subcommands when I need to do something short of "write a java program".  There are a few issues:

 * The various messages like "Creating connection configuration" make it hard to get at the exact byte-for-byte data.
 * Loading everything into memory doesn't work for large files.
 * no way to just write to a file (a la `mget -O`)
 * I'm not sure `getAsString` is going to do what we want for raw binary data.

dekobon:
@geek What is the status on the node support for CSE?

geek:
@dekobon it works fine in my branch but is waiting on approval for merging into the mainline

cburroughs:
Still https://cr.joyent.us/#/c/1110/ ?

PR #329, opened 2017-08-31
make cli: get-file suitable for streaming

cburroughs:
Instead of reading the entire object into a string and running out of
memory, write out bytes as they are downloaded.  This also makes it
easy to stream binary content to a local file.

Add a nascent info command to retain a way for the cli to show
http headers.

ref #312

### DIFF
--- a/java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
+++ b/java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
@@ -18,12 +18,17 @@ import com.joyent.manta.config.ConfigContext;
 import com.joyent.manta.config.DefaultsConfigContext;
 import com.joyent.manta.config.EnvVarConfigContext;
 import com.joyent.manta.config.MapConfigContext;
+import com.joyent.manta.util.MantaUtils;;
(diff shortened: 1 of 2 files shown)


java-manta-cli/src/main/java/com/joyent/manta/cli/MantaCLI.java
```diff
@@ -264,33 +270,47 @@ public void run() {
...
+                    out = new FileOutputStream(outputFileName);
+                }
+                BufferedOutputStream bout = new BufferedOutputStream(out, 1024);
 
-                b.append("Attempting GET request to: ").append(filePath).append(BR);
-                b.append("\nMetadata values: \n").append(client.get(filePath).getMetadata().toString());
-                b.append("\n\nPayload: \n");
+                InputStream objectStream = client.getAsInputStream(filePath);
+                int oneByte;
+                while ((oneByte = objectStream.read()) != -1) {
```
tjcelaya: Any particular reason to reimplement the `IOUtils.copy` loop for a single byte at a time?

cburroughs: Accident of development process, no good reason.

Issue #332, opened 2017-09-14
EncryptedServerSideMultipartManagerIT#canRetryUploadPart intermittently fails unexpectedly


tjcelaya:
Under certain test scenarios it is possible for the `EncryptedServerSideMultipartManagerIT#canRetryUploadPart` test case to fail during commit. The thrown exception looks like the following:

```
com.joyent.manta.exception.MantaMultipartException: Unable to create multipart upload
Exception Context:
	[1:mantaSdkVersion=3.0.0-SNAPSHOT]
	[2:requestId=c35ed0af-98db-11e7-9489-936727b412d6]
        ...
	[10:responseStatusCode=409]
	[11:responseStatusReason=Conflict]
	[12:objectPath=/java.manta.ci/stor/java-manta-integration-tests/4befbbe4-17b6-4ff4-be03-b0d03b4acd17/b064f1a9-2919-4db2-8cf4-8b191d8beb48]
	[13:requestBody={"parts":["9926b93e-2d94-6575-a8d5-fa1a3c16f411","1588463a-2645-e07c-c498-9b0039c3d0dc","5bdf30e0-aeea-eda0-b276-8d0ffea38d61"]}]
---------------------------------

	at com.joyent.manta.client.multipart.ServerSideMultipartManager.validateStatusCode(ServerSideMultipartManager.java:730)
	at com.joyent.manta.client.multipart.ServerSideMultipartManager.complete(ServerSideMultipartManager.java:623)
	at com.joyent.manta.client.multipart.ServerSideMultipartManager.complete(ServerSideMultipartManager.java:68)
	at com.joyent.manta.client.multipart.EncryptedMultipartManager.complete(EncryptedMultipartManager.java:334)
	at com.joyent.manta.client.multipart.EncryptedServerSideMultipartManagerIT.canRetryUploadPart(EncryptedServerSideMultipartManagerIT.java:569)
	...

Caused by: com.joyent.manta.exception.MantaClientHttpResponseException: HTTP request failed to: /java.manta.ci/stor/java-manta-integration-tests/4befbbe4-17b6-4ff4-be03-b0d03b4acd17/b064f1a9-2919-4db2-8cf4-8b191d8beb48
Exception Context:
	[1:mantaSdkVersion=3.0.0-SNAPSHOT]
	[2:statusLine=HTTP/1.1 409 Conflict]
	[3:serverCode=MultipartUploadInvalidArgument]
[...]

PR #296, opened 2017-08-02
Fix #295

tjcelaya:
See #295

PR #333, opened 2017-09-15
EncryptingEntity and EncryptedPartEntity should not close writeTo(OutputStream)

tjcelaya:
This PR adds test cases for `EncryptingEntity` and an integration test for `EncryptingPartEntity`. 

Apologies for the large segment of changes in `EncryptingEntityTest.java` I noticed the last test case I previously added (`canBeWrittenIdempotently`) was after the utility methods so I moved it to the right place before adding `doesNotCloseSuppliedOutputStreamWhenFailureOccurs` and `doesNotCloseSuppliedOutputStreamWhenWrittenSuccessfully`.

These entities should be writing content to the provided `OutputStream` but *not* closing it, otherwise requests using chunked encoding (those backed by an `InputStream`) complete successfully even if they were intended to fail. This results in intermittent test failures during the MPU commit operation where the intentionally-failing payloads in `EncryptedServerSideMultipartManagerIT#canRetryUploadPart` can "win out" over the full part content. This is visible in the following directory listing of an upload directory for a commit which failed on 409 Conflict as described in #332:

```
mls -l ~~/uploads/2/20c5f2f5-efe3-cc72-fe73-e89a7a4bc123
-rwxr-xr-x 1 java.manta.ci          1024 Sep 11 15:30 0
-rwxr-xr-x 1 java.manta.ci       5242880 Sep 11 15:30 1
-rwxr-xr-x 1 java.manta.ci           512 Sep 11 15:30 2
```

The `1024` byte part 0 would not have been created if `EncryptingPartEntity` had left the `OutputStream` intact (so that HttpClient could terminate the request correctly).

### DIFF
--- a/java-manta-client-unshaded/src/test/java/com/joyent/manta/client/crypto/EncryptingEntityTest.java
+++ b/java-manta-client-unshaded/src/test/java/com/joyent/manta/client/crypto/EncryptingEntityTest.java
@@ -172,7 +172,7 @@ public class EncryptingEntityTest {
         }
     }
 
-    public void doesNotCloseSuppliedOutputStreamWhenFailureOccurs() throws Exception {
+    public void doesNotCloseSuppliedOutputStreamWhenWrittenSuccessfully() throws Exception {
         final SupportedCipherDetails cipherDetails = DefaultsConfigContext.DEFAULT_CIPHER;
         final SecretKey secretKey = SecretKeyUtils.generate(cipherDetails);
         final String content = RandomStringUtils.randomAlphanumeric(RandomUtils.nextInt(500, 1500));
@@ -190,7 +190,7 @@ public class EncryptingEntityTest {
         Mockito.verify(output, Mockito.never()).close();
     }
 
-    public void doesNotCloseSuppliedOutputStreamWhenWrittenSuccessfully() throws Exception {
+    public void doesNotCloseSuppliedOutputStreamWhenFailureOccurs() throws Exception {
         final SupportedCipherDetails cipherDetails = DefaultsConfigContext.DEFAULT_CIPHER;
         final SecretKey secretKey = SecretKeyUtils.generate(cipherDetails);
         final EncryptingEntity encryptingEntity = new EncryptingEntity(
--- a/java-manta-client-unshaded/src/test/java/com/joyent/manta/client/crypto/EncryptingEntityTest.java
+++ b/java-manta-client-unshaded/src/test/java/com/joyent/manta/client/crypto/EncryptingEntityTest.java
@@ -9,6 +9,7 @@ package com.joyent.manta.client.crypto;
 
 import com.joyent.manta.client.MantaObjectInputStream;
 import com.joyent.manta.client.MantaObjectResponse;
+import com.joyent.manta.config.DefaultsConfigContext;
 import com.joyent.manta.exception.MantaClientEncryptionException;
 import com.joyent.manta.http.MantaHttpHeaders;
 import com.joyent.manta.http.entity.ExposedStringEntity;
@@ -17,11 +18,13 @@ import com.joyent.manta.util.FailingOutputStream;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.input.BoundedInputStream;
+import org.apache.commons.io.input.BrokenInputStream;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.RandomUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.conn.EofSensorInputStream;
+import org.apache.http.entity.InputStreamEntity;
 import org.bouncycastle.jcajce.io.CipherInputStream;
 import org.mockito.Mockito;
 import org.testng.Assert;
@@ -34,6 +37,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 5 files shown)


tjcelaya:
(This is the client-side resolution for MANTA-3423, server-side investigation still pending)

cburroughs:
If `HttpEntity.writeTo` isn't supposed to close (the [javadocs](https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/HttpEntity.html#writeTo-java.io.OutputStream-) are sadly not clarifying), do we know what will do the close?

Do we know why (aka what code) looked at the outputstream and said "this was closed so everything is good and complete) which caused the original issue?

tjcelaya:
The correct place for the provided `OutputStream` to be closed is in [DefaultBHttpClientConnection#sendRequestEntity](https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/impl/DefaultBHttpClientConnection.java#L147) which builds the `OutputStream`, calls `HttpEntity#writeTo` with it and then calls `close` on the stream. In the face of network or input errors, an exception would be thrown which prevents the regular call to `close`, but the catch causes in `MainClientExec#execute` are supposed to [abort the connection](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/execchain/MainClientExec.java#L344) before they rethrow. In the case of `ChunkedOutputStream`, calling `close` when an error occurs is *plain wrong* because it insinuates that the request terminated successfully.

java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerIT.java
```diff
@@ -576,4 +578,27 @@ public final void canRetryUploadPart() throws IOException {
...
+        EncryptedMultipartUpload<ServerSideMultipartUpload> upload = multipart.initiateUpload(path);
+
+        Assert.assertThrows(IOException.class, () -> {
+            // partial read of content1
+            multipart.uploadPart(upload, 1, new FailingInputStream(new ByteArrayInputStream(content), 1024));
+        });
+
+        // listing the directory immediately after MPU creation is likely to return no results
+        // this is being reported upstream
+        TimeUnit.SECONDS.sleep(1);
```
dekobon: Shouldn't you do a loop around the sleep in case your sleep wait value is too small?

tjcelaya: This is a really weird context to try to explain..

Based on the fact that you need to see a test fail before it passes I was trying to design the test so it would fail without the proposed change (during which I temporarily reverted the fix). While running the test case I discovered that the uploaded part does not appear in the MPU part directory immediately after the PUT request completes, so the test case was passing both *with* and *without* the change. 

The [RFD](https://github.com/joyent/rfd/blob/master/rfd/0065/README.md#2-upload-part) seems to imply that there might be a race condition between upload-part and MPU state. I couldn't find anything that might imply directory listings would be affected though.

In summary, we don't really have anything to check in a loop since the expectation is that the part never appears. This test is starting to seem a bit contrived, I'll need to rethink whether it's worth keeping. Perhaps it would make more sense to create a unit test for `EncryptingPartEntity` like the test cases created for `EncryptingEntity`?

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/multipart/EncryptingPartEntityTest.java
```diff
@@ -0,0 +1,81 @@
+package com.joyent.manta.client.multipart;
```
cburroughs: ^^ License headers

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/multipart/EncryptingPartEntityTest.java
```diff
@@ -0,0 +1,81 @@
...
+                state.getMultipartStream(),
+                new InputStreamEntity(new BrokenInputStream(new IOException("bad input"))),
+                CALLBACK_NOOP);
+
+        final OutputStream output = Mockito.mock(OutputStream.class);
+        Assert.assertThrows(IOException.class, () -> encryptingPartEntity.writeTo(output));
+        Mockito.verify(output, Mockito.never()).close();
+    }
+
+}
```
cburroughs: nit: newline

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/crypto/EncryptingEntityTest.java
```diff
@@ -116,6 +120,89 @@ public void canSurviveNetworkFailuresInAesCbc() throws Exception {
...
+
+    private void canBeWrittenIdempotently(final SupportedCipherDetails cipherDetails) throws Exception {
+        final SecretKey secretKey = SecretKeyUtils.generate(cipherDetails);
+        final String content = RandomStringUtils.randomAlphanumeric(RandomUtils.nextInt(500, 1500));
+        final ExposedStringEntity contentEntity = new ExposedStringEntity(
+                content,
+                StandardCharsets.UTF_8);
+
+        final ByteArrayOutputStream referenceEncrypted = new ByteArrayOutputStream(content.length());
+        {
```
cburroughs: I'm not following what this bracket signifies.

tjcelaya: This test was introduced with #296 and moved in this PR when I noticed it was below the test utility methods. The reason for the bracket was to isolate variable scopes, by the time I wrote this test I had been bitten by variable naming that made some tests seem to pass when they were supposed to fail and vice-versa. The braces made mistakenly-reused variables from the first round of encryption into compiler errors so I could be sure I wasn't mixing objects between encryption operations.

Issue #331, opened 2017-09-04
Document Behavior of `MantaClient#close`


tjcelaya:
The current behavior of `MantaClient#close` is to traverse the `danglingStreams` collection and close potentially-open stream in addition to shutthing down the connection pool but is not explicit about multi-threaded behavior (it does not provide a JavaDoc). Depending on how users share instances between threads, they may see one of the following exceptions as a result of the client being shut down in a separate thread:

- `java.lang.IllegalStateException: Connection pool shut down`
- `java.io.IOException: Attempted read on closed stream`

This leads to two questions:
- How do we signal to other threads that the client is shutting down?
  - Conversely, should the client throw an exception on `close()` to indicate that there are still pending requests? This might be a reasonable improvement over our current behavior where we block indefinitely until all resources are successfully closed and could be provided as a secondary close method similar to the `awaitTermination(long timeout, TimeUnit unit)` method of `ExecutorService`
- Should every method exposed by the client check if the client is closed before proceeding? This seems like overkill since HttpClient will eventually throw `IllegalStateException: Connection pool shut down`

At the very least, adding a JavaDoc to `MantaClient#close` about its behavior would be beneficial for users.

PR #334, opened 2017-09-16
Proposal for potentially faster MantaClient#close

tjcelaya:
This PR serves to push forward discussion of #331. The proposal is to add a configuration option `manta.fast_close`/`MANTA_FAST_CLOSE` which applies the pattern used for `danglingStreams` to keep track of in-flight requests in order to abort them. I've created a test case utilizing a stub `com.sun.net.httpserver.HttpServer` which is designed to hold connections open indefinitely in order to minimize timing issues. The test case then creates a `CountDownLatch` and spawns three threads:

 - downloader, calls `getAsInputStream` and decrements the latch
 - uploader, calls `put` and passes a `ProxyInputStream` which will decrement the latch after the first `read`
 - terminator, decrements the latch, waits for the other threads to be ready and closes the `MantaClient` object on unlatch

After spawning those threads, the test case waits on the latch to trigger shutdown of the thread pool. By using the return value of  `ExecutorService#awaitTermination` we can tell whether or not the terminator thread was able to trigger exceptions in the download/uploader threads. If the return value is true the thread pool terminated successfully, otherwise the main thread will forcibly terminate the thread pool, make sure that the `MantaClient` is closed, and fail on any unexpected exceptions the worker threads may have encountered.

After revisiting this PR and ripping out `FutureRequestExecutorService` in favor of checking for aborted requests in-line. Calls to `HttpClient#execute` will now wrap the thrown exception in  `MantaRequestAbortedException` based on `request.isAborted()`.


tjcelaya:
One thing to note is the presence of a Boolean on some requests to check if it had been aborted.

Perhaps catching and rethrowing if `request.isAborted() == true` could be used to wrap whatever exception is actually throw in a `MantaInterruptedException` making this usage easier for callers to discern.

tjcelaya:
This was an interesting experiment but seems very fragile. I get the sense this feature doesn't fit into the current abstraction very well, but would be trivial when using AsyncHttpClient

PR #335, opened 2017-09-18
Actually disable automatic retries when set in ConfigContext

tjcelaya:
Unless explicitly disabled, automatic retries will be configured by `HttpClientBuilder`

https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/client/HttpClientBuilder.java#L1128


cburroughs:
Is this bug user visible in a way that should be called out in the changelog?

tjcelaya:
I believe most users would not be affected by the bug but it's possible for someone to notice this behavior so I opted to call it out in the changelog.

Issue #320, opened 2017-08-25
would like integration tests to use a dedicated per-run directory


cburroughs:
Currently we place all of the integration tests objects under `~~/stor/java-manta-integration-tests` with UUIDs to ensure unique object names.  All objects should be cleaned  up after tests complete, but due to test failures, bugs in the tests themselves (#319), job cancellations, etc this is not always the case.

In one environment, I would like to be able to `mrm -r ~~/stor/java-manta-integration-tests` because the test account is using an appreciable amount of sapce, but there are a few objects I need for ongoing debugging.  It would be easier to find the files I need, and to delete everything else, if each invocation of the integration tests used it's own directory (ie `~~/stor/java-manta-integration-tests/$UUID`).

tjcelaya:
How is this different from #319 ?

cburroughs:
#319 is for  `~~/stor/java-manta-integration-tests/$UUID_PER_CLASS` not being used correctly.

This proposes changing to `~~/stor/java-manta-integration-tests/$UUID_PER_RUN/$UUID_PER_CLASS`.  Although as I look at that it might be nice to change from `$UUID_PER_CLASS` to `$PackageName`, since we would then only need to be unique per run.

tjcelaya:
Sorry I missed the distinction. The PR for #319 centralizes path generation
so it should pave the way for addressing this issue. I'd actually thought
about proposing a change like this one at one point so I'm definitely on
board.

Issue #180, opened 2017-02-04
integration test cleanup should be controlled by a parameter 


cburroughs:
Many of the integration tests will `deleteRecursive` or otherwise wipe all remote evidence of their run after they finish.  When investigating failures, it would be nice to preserve data without commenting out code.

dekobon:
Why not set a breakpoint at the end of a test? A parameter would be fine too.

PR #337, opened 2017-09-20
per-run integration test directories with cleanup param

cburroughs:
Move integration test files from
  ~~/stor/java-manta-integration-tests/$UUID_PER_CLASS/
 to
  ~~/stor/java-manta-integration-tests/$UUID_PER_RUN/$CLASS_NAME/

This makes it easier for some manual debugging, and managing data
retention after many integration test runs.  In centralizing cleanup
it was also convenient added a way to *not* cleanup after test runs.

ref #180 #320


cburroughs:
(Fixed up the name difference before landing.)

PR #344, opened 2017-09-26
Extract MantaHttpRequestFactory from MantaConnectionFactory

tjcelaya:
This PR is the first part of a larger branch that works towards #310 and #311. This part extracts the convenience http request creation convenience methods from `MantaConnectionFactory` which is already responsible for configuring connection parameters and registering MBeans.

Apologies for the dull PR, most of the lines are replacements of `connectionFactory` -> `requestFactory`

Issue #311, opened 2017-08-16
Lazily read ConfigContext in MantaClient


dkocher:
We would want to provide connection details after instantiation of the client and allow to change configuration options such as the `mantaKeyId` afterwards.

dekobon:
The `MantaClient` class was designed so that each instance's runtime configuration is immutable. If you wanted to change the configuration, you would typically close an instance and then create another one.

If we were to introduce mutability into its state, then it may be possible by swapping out the `HttpHelper` and the `ConfigContext` instances. However, that would be very difficult to reasonable in a multi-threaded environment (which is very typical of a lot of users of Manta).

Imagine that you have instantiated a MantaClient, you now have passed it off to 4 threads that are using it to make concurrent requests to Manta. Now, I would like to change the runtime configuration of that class. How do I signal to the threads that they need to stop their operations so that I can destroy and recreate the connection pool / other runtime settings?

I wonder if we can introduce an abstraction that gives you an elegant way to integrate with the design of your application without adding complexity? Does anyone have any ideas?

tjcelaya:
I have a `ReconfigurableMantaClient` sitting in a branch somewhere which acts as a proxy to an actual `MantaClient` and would `close` and discard the previous client before building a new one. The actual implementation was ugly (`MantaClient`'s methods were copied into an interface and implemented by both) so I was hesitant to propose it as a solution.

Alternatively, "closing" the `MantaClient` could be made more flexible by acting as a toggle instead of a one-off operation. Having the constructor delegate to a public `configure(ConfigContext)` method which grabs a lock and calls `close` as its first action would keep the configuration/thread-management lifecycle in the `MantaClient` class and avoid the creation of multiple unnecessary `MantaClient` objects.

In the end I could not decide between these two approaches but those are my suggestions.

dekobon:
Both of those solutions seem viable to me.

PR #352, opened 2017-10-04
Allow users to supply a custom HttpClientBuilder and HttpClientConnectionManager

### LINKED ISSUE #310: Allow configuration of client with custom MantaApacheHttpClientContext
Allow to configure `MantaClient` with a custom context. We want to use our own configured `HttpClient` instance.

tjcelaya:
Resolves #310. 

## Purpose
Accept an `HttpClientBuilder` and associated `HttpClientConnectionManager` (since the connection manager handles SSL configuration) and allow `MantaConnectionFactory` to work with both provided and internally-constructed versions of both.

## Why create a new class?
~These two objects are closely coupled and one has a lifecycle that needs to be managed. I'm open to switching to passing the builder and manager directly but this way allows us to avoid null checks.~ Keeping the class for use in #357

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -247,7 +247,6 @@ public class MantaClient implements AutoCloseable {
         if (BooleanUtils.isTrue(config.isClientEncryptionEnabled())) {
             this.httpHelper = new EncryptionHttpHelper(connectionContext, config);
         } else {
-
             this.httpHelper = new StandardHttpHelper(connectionContext, config);
         }
 
@@ -2406,7 +2405,6 @@ public class MantaClient implements AutoCloseable {
         danglingStreams.add(stream);
         return stream;
     }
-
 
     /* ======================================================================
      * Lifecyle Methods
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -28,6 +28,7 @@ import com.joyent.manta.http.EncryptionHttpHelper;
 import com.joyent.manta.http.HttpHelper;
 import com.joyent.manta.http.MantaApacheHttpClientContext;
 import com.joyent.manta.http.MantaConnectionFactory;
+import com.joyent.manta.http.MantaConnectionFactoryConfigurator;
 import com.joyent.manta.http.MantaContentTypes;
 import com.joyent.manta.http.MantaHttpHeaders;
 import com.joyent.manta.http.StandardHttpHelper;
@@ -236,18 +237,14 @@ public class MantaClient implements AutoCloseable {
     }
 
     /**
-     * Creates a new instance of the Manta client with the specified connection factory.
-     * This method is typically used by unit tests to inject connection factory
-     * customizations so we can simulate different network events.
+     * Creates a new instance of the Manta client based on user-provided connection objects. This allows for a larger
+     * degree of customization at the cost of
      *
      * @param config The configuration context that provides all of the configuration values
-     * @param keyPair cryptographic signing key pair used for HTTP signatures
-     * @param connectionFactory connection factory instance used to tune connection settings
+     * @param connectionFactoryConfigurator pre-configured objects for use with a MantaConnectionFactory
      */
-    MantaClient(final ConfigContext config,
-                final KeyPair keyPair,
-                final MantaConnectionFactory connectionFactory,
-                final ThreadLocalSigner signer) {
+    public MantaClient(final ConfigContext config,
+                       final MantaConnectionFactoryConfigurator connectionFactoryConfigurator) {
         dumpConfig(config);
 
         ConfigContext.validate(config);
@@ -256,10 +253,24 @@ public class MantaClient implements AutoCloseable {
         this.config = config;
         this.home = ConfigContext.deriveHomeDirectoryFromUser(config.getMantaUser());
 
+        final KeyPair keyPair = new KeyPairFactory(config).createKeyPair();
+
+        final Signer.Builder builder = new Signer.Builder(keyPair);
+        if (ObjectUtils.firstNonNull(
+                config.disableNativeSignatures(),
+                DefaultsConfigContext.DEFAULT_DISABLE_NATIVE_SIGNATURES)) {
+            builder.providerCode("stdlib");
+        }
+        final ThreadLocalSigner signer = new ThreadLocalSigner(builder);
         this.signerRef = new WeakReference<>(signer);
 
-        final MantaApacheHttpClientContext connectionContext =
-                new MantaApacheHttpClientContext(connectionFactory);
+        final MantaConnectionFactory connectionFactory = new MantaConnectionFactory(
+                config,
+                keyPair,
+                signer,
+                connectionFactoryConfigurator);
+
+        final MantaApacheHttpClientContext connectionContext = new MantaApacheHttpClientContext(connectionFactory);
 
         if (BooleanUtils.isTrue(config.isClientEncryptionEnabled())) {
             this.httpHelper = new EncryptionHttpHelper(connectionContext, config);
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -209,7 +209,7 @@ public class MantaClient implements AutoCloseable {
      *
      * Users opting into advanced configuration (i.e. not passing {@code null} as the second parameter)
      * should be comfortable with the internals of {@link CloseableHttpClient} and accept that we can only make a
-     * best effort to support all possible use-cases. For example, uses may pass in a
+     * best effort to support all possible use-cases. For example, uses may pass in a builder which is wired to a
      * {@link org.apache.http.impl.conn.BasicHttpClientConnectionManager} and effectively make the client
      * single-threaded by eliminating the connection pool. Bug or feature? You decide!
      *
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
@@ -39,7 +39,6 @@ import com.joyent.manta.util.MantaUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.BooleanUtils;
-import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.Validate;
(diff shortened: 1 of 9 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
```diff
@@ -341,16 +376,6 @@ public CloseableHttpClient createConnection() {
...
 
-    /**
-     * package-private method for building a {@link MantaHttpRequestFactory} from this object's
-     * config. Should be removed with the deprecated constructor for {@link MantaApacheHttpClientContext}.
-     *
-     * @return a request factory pointed at the same url as {@code this}
-     */
-    ConfigContext getConfig() {
-        return config;
-    }
```
tjcelaya: Removing unused method that was mistakenly included by #344

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -236,18 +237,14 @@ public MantaClient(final ConfigContext config) {
     }
 
     /**
-     * Creates a new instance of the Manta client with the specified connection factory.
-     * This method is typically used by unit tests to inject connection factory
-     * customizations so we can simulate different network events.
+     * Creates a new instance of the Manta client based on user-provided connection objects. This allows for a larger
+     * degree of customization at the cost of
```
cburroughs: Trailing off?

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
```diff
@@ -86,10 +88,11 @@
     /**
      * Default HTTP headers to send to all requests to Manta.
      */
-    private static final Collection<? extends Header> HEADERS = Arrays.asList(
-            new BasicHeader(MantaHttpHeaders.ACCEPT_VERSION, "~1.0"),
-            new BasicHeader(HttpHeaders.ACCEPT, "application/json, */*")
-    );
+    public static final Collection<? extends Header> HEADERS =
+            Collections.unmodifiableList(
```
cburroughs: Is this needed for this change?  It looks like a fine improvement but was unsure if it was demonstrating one of the "spooky"  issues with all this being user configurable.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -256,10 +253,24 @@ public MantaClient(final ConfigContext config) {
...
 
+        final KeyPair keyPair = new KeyPairFactory(config).createKeyPair();
+
+        final Signer.Builder builder = new Signer.Builder(keyPair);
+        if (ObjectUtils.firstNonNull(
+                config.disableNativeSignatures(),
+                DefaultsConfigContext.DEFAULT_DISABLE_NATIVE_SIGNATURES)) {
+            builder.providerCode("stdlib");
+        }
+        final ThreadLocalSigner signer = new ThreadLocalSigner(builder);
```
cburroughs: This block looks the same as in the other constructor.  Should it be DRYed out?

I probably wrote this block,but I admit I am having trouble recalling why `DefaultsConfigContext` needs to explicitly be referenced?  We allow config contexts that don't start with the defaults?

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -236,18 +237,14 @@ public MantaClient(final ConfigContext config) {
...
      *
      * @param config The configuration context that provides all of the configuration values
-     * @param keyPair cryptographic signing key pair used for HTTP signatures
-     * @param connectionFactory connection factory instance used to tune connection settings
+     * @param connectionFactoryConfigurator pre-configured objects for use with a MantaConnectionFactory
      */
-    MantaClient(final ConfigContext config,
-                final KeyPair keyPair,
-                final MantaConnectionFactory connectionFactory,
-                final ThreadLocalSigner signer) {
```
cburroughs: Dropping the ability to pass in a signer is intentional, correct?

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/MantaClientConnectionFailuresIT.java
```diff
@@ -112,13 +112,20 @@ public NoHttpResponseMantaConnectionFactory(ConfigContext config, KeyPair keyPai
         }
     }
 
+    /**
+     * Test is disabled in testng.xml, the MantaClient constructor needs to be revisted.
+     *
+     * @throws IOException
+     */
+    @Test(enabled = false)
```
cburroughs: Is the intention here to fix this resolve this before merging?

java-manta-client-unshaded/src/test/java/com/joyent/manta/http/MantaConnectionFactoryConfiguratorTest.java
```diff
@@ -0,0 +1,33 @@
+package com.joyent.manta.http;
```
cburroughs: ^^ License headers

java-manta-client-unshaded/src/test/java/com/joyent/manta/http/MantaConnectionFactoryTest.java
```diff
@@ -0,0 +1,197 @@
+package com.joyent.manta.http;
```
cburroughs: ^^ License headers.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
tjcelaya: This constructor was only used in a single test which was has been disabled for a while (by being commented out in the `testng.xml` file) and needs to be revisted. That test case is documented in #354 and could be achieved with the `MantaConnectionFactoryConfigurator`  and some simple mocks.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
tjcelaya: Must've been interrupted, thanks for catching that.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
tjcelaya: I remember punting on the decision between:

- document default headers and tell `MantaConnectionFactoryConfigurator` users to include these headers in their `HttpClientBuilder` configuration
- add these headers to whatever default headers are provided by the user. This might require inspecting the `HttpClientBuilder`'s internal state to determine if we've already added them in #311?
- add a `setDefaultHeaders` method to `MantaConnectionFactoryConfigurator` so users can tell us what headers they want to mix into the `HttpClientBuilder`
- something else?

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/MantaClientConnectionFailuresIT.java
tjcelaya: No, this test has been disabled since [December 2016](https://github.com/joyent/java-manta/commit/49c0e5d5a1f993b5e86907a24b60ef61b42da875) and has been documented in #354 to be addressed soon (but not here).

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
tjcelaya: People could pass a `StandardConfigContext` which would throw NPEs if accessed directly. I've been fighting with the tradeoff between documenting defaults by directly referencing the `DefaultsConfigContext` and expressing the default behavior through for example structuring if clauses and using `BooleanUtils.isTrue()` for handling nulls. This small check would probably benefit from the latter since it's a simple boolean check though.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
dekobon: One thought is to change the `HEADERS` constant into an array of `org.apache.http.Header[]` objects and store it as a constant in `MantaHttpRequestFactory` or in `StandardHttpHelper` and add the headers to every `HttpHead`, `HttpGet`, etc. This would be a solution that would always make sure the right headers were there regardless of the `HttpClientBuilder` that was used.

tjcelaya: That suggestion led me to discover default headers are implemented as just another `HttpRequestInterceptor`, definitely going to use `MantaHttpRequestFactory` for that and remove our call to `HttpClientBuilder#setDefaultHeaders`. Thanks!

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactoryConfigurator.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * An existing {@link HttpClientBuilder} to further configure.
+     */
+    private final HttpClientBuilder httpClientBuilder;
+
+    /**
+     * An existing {@link HttpClientConnectionManager} to use with the {@link #httpClientBuilder}.
+     */
+    private final HttpClientConnectionManager connectionManager;
```
dkocher: This parameter could be omitted and set to `null` in `MantaConnectionFactory`.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
```diff
@@ -116,38 +117,49 @@
...
-            Validate.notNull(signer, "Signer must not be null if authentication is enabled");
-
-            // pass true directly to the constructor because auth is enabled
-            final HttpRequestInterceptor authInterceptor = new HttpSignatureRequestInterceptor(
-                            new HttpSignatureAuthScheme(keyPair, signer),
-                            new UsernamePasswordCredentials(config.getMantaUser(), null),
-                            true);
-            this.httpClientBuilder.addInterceptorLast(authInterceptor);
+        if (connectionFactoryConfigurator != null) {
+            this.connectionManager = connectionFactoryConfigurator.getConnectionManager();
```
dkocher: Set to `null` instead.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
```diff
@@ -116,38 +117,49 @@
     private final HttpClientConnectionManager connectionManager;
 
     /**
+     * Flag indicating if we created the {@link #connectionManager} ourselves and are responsible for shutting it down.
+     */
+    private final boolean connectionManagerShared;
```
dkocher: Remove unused variable as null check will suffice.

Issue #315, opened 2017-08-17
ConfigMBean Allows Configuration Drift from MantaClient


tjcelaya:
While doing some investigation into #311 I noticed we allow editing of quite a few properties through `ConfigMBean`. As an example of a potential issue, it's currently possible to construct a `MantaClient` with one `manta.key_path` value and then change that path to point to a different key without impacting the key actually used for signing requests. A simpler example might be the `home` value used to list running jobs which is only set on construction of a `MantaClient`. The expected fix for this issue is at the bare minimum to make the attributes read-only through the MBean but support for tweaking values interactively (like `manta.max_connections`) and their documentation would be nice.

dekobon:
Typically, access to MBeans is limited to operators and developers. The only time anyone should be muckering with those values is when doing performance testing. I don't see any problem with changing them to read-only. However, it might be a good way to experiment with a mutable version of a `Context`.

cburroughs:
A user might assume that anything implementing `SettableConfigContext` already is a mutable `ConfigContext`.  Maybe the javadoc should could that?  `ConfigContextBuilder` might have been a name more in line with intended use.

PR #357, opened 2017-10-06
Lazy-loading configuration for MantaClient

tjcelaya:
This is a proposal for a refactoring of `MantaClient` and its helper classes which can respond to configuration changes as described in #311. `AuthAwareConfigContext` is the primary focus for handling dynamic config updates of the objects we shouldn't be building on-the-fly per-request.

~(This PR depends on #352 and will be rebased once that is merged. It is bested [viewed here](https://github.com/tjcelaya/java-manta/compare/enhancement/310-supply-custom-HttpClientBuilder...tjcelaya:enhancement/311-reconfigure-mantaclient) until that is merged.)~


tjcelaya:
# Summary
In order to allow for the `MantaClient` to accommodate changes to it's underlying `ConfigContext` without duplicating all of its object access methods (or migrating them to an external "operator" object), I've replaced direct property access in `MantaClient` to use getters instead. Since most users will be using the regular `MantaClient` which has these fields marked `final` the getter _should_ easily be JIT'd into regular field access.

## Detecting changes

An `Observable` is definitely an option but I wanted to keep it simple for this PR. Asking users to call `reload()` on the client after modifying the `ConfigContext` doesn't seem _too_ egregious, but it could definitely be better. Is it too clunky? Is there something else we can do that isn't overly complex.

## Responding to change

The `EnumSet<MantaClientComponent>` property keeps track of which components have been refreshed. Calls to `reload()` fill this set with all values of the `MantaClientComponent` enum and getters will trigger cleanup and replacement of relevant fields.

`AuthenticationConfigurator` acts similarly but takes an all-or-nothing approach since the fields it's interested in are coupled to most of it's derived objects. Since non-authentication-related parameters can be changed in the `ConfigContext` a "fingerprint" of the last params used is kept in order to skip reload if nothing relevant has changed.

## The #352 Problem

This feature combines with #352 to cause what is effectively a memory leak unless properly mitigated. Since the configuration of externally-provided `HttpClientBuilder` objects is mutable and we attach `RequestIdInterceptor` and `HttpSignatureAuthenticationInterceptor` objects to the lists inside the `HttpClientBuilder` it is possible for those lists to grow without bound and not notice since the interceptors are deduplicated when the `CloseableHttpClient` is created. This could _also_ affect correctness if a user were to disable authentication completely since the previously-attached `HttpSignatureAuthenticationInterceptor` actually be used when the client is rebuilt. On the other hand we don't want to deep-clone everything in `HttpClientBuilder` because it could contain lambas (of which we can't generally instantiate new copies) attached as request handlers, deep object hierarchies in custom  `Registry` or manager objects, etc. See the [test cases for `HttpClientBuilderShallowCloner`](https://github.com/joyent/java-manta/pull/357/files#diff-5ffbd28a9e573427940c34cf16dd2bb5),
 [`LazyMantaClientTest#canReloadHttpHelperRepeatedlyWithoutLeakingInterceptors`](https://github.com/joyent/java-manta/pull/357/files#diff-1310e9b2f7ad8583882de1889762ad49R163), and [this docblock in `MantaConnectionFactoryConfigurator`](https://github.com/joyent/java-manta/pull/357/files#diff-7d037ee5cbaa58c03c16c8a70d29c1c6R24) for more context.

# Design questions

The `LazyMantaClient` class leverages getters in `MantaClient` to achieve its magic without requiring an class just to actually execute the requests. Alternatives evaluated (with attached drawbacks indented further) include:
- An "operator" class which contains the bodies of the object access methods in `MantaClient` but accepts instances of `HttpHelper`, `UriSigner`, etc. for each method call 
  - This would leave the `MantaClient` class with a ton of methods that just invoke command objects or something to that effect and make the central class in the library more abstract than I'd like
- An interface for the barebones `put`, `get`, `getAsInputStream` methods which are buried in `MantaClient` alongside their convenience overloads
   - This would lead to a lot of extra refactoring to pull out a minimal interface for what the `MantaClient` does and would probably still lead to code duplication.
- An `abstract` base MantaClient which has none of its own fields and just uses getters everywhere.
  - All of the non-initialization code in `MantaClient` would move to this class and the diff would be much larger
  - **This might actually be the best option and is a simple transition from where we are now, but I want someone else to weigh in on whether this is actually valuable**

The current implementation seems suboptimal since we're extending a parent class and then setting all of it's fields to `null`, but I wanted to avoid removing the `final` modifier from `MantaClient`'s fields because they truly are immutable in `MantaClient`. As far as I can tell the impact is twofold:
 - that there are extra fields in `LazyMantaClient` taking up space
 - maintainers might be confused why the fields aren't just mutable in `MantaClient`
The object layout gives a better idea of what's going on:
```
com.joyent.manta.client.MantaClient object internals:
 OFFSET  SIZE                                                       TYPE DESCRIPTION                               VALUE
      0    12                                                            (object header)                           N/A
     12     1                                                    boolean MantaClient.closed                        N/A
     13     3                                                            (alignment/padding gap)                  
     16     4                                           java.lang.String MantaClient.url                           N/A
     20     4                           com.joyent.manta.http.HttpHelper MantaClient.httpHelper                    N/A
     24     4                                           java.lang.String MantaClient.home                          N/A
     28     4                      com.joyent.manta.config.ConfigContext MantaClient.config                        N/A
     32     4                                              java.util.Set MantaClient.danglingStreams               N/A
     36     4                                java.lang.ref.WeakReference MantaClient.signerRef                     N/A
     40     4                          com.joyent.manta.client.UriSigner MantaClient.uriSigner                     N/A
     44     4               com.joyent.manta.client.MantaMBeanSupervisor MantaClient.beanSupervisor                N/A
Instance size: 48 bytes
Space losses: 3 bytes internal + 0 bytes external = 3 bytes total

--------
com.joyent.manta.client.LazyMantaClient object internals:
 OFFSET  SIZE                                                       TYPE DESCRIPTION                               VALUE
      0    12                                                            (object header)                           N/A
     12     1                                                    boolean MantaClient.closed                        N/A
     13     3                                                            (alignment/padding gap)                  
     16     4                                           java.lang.String MantaClient.url                           N/A
     20     4                           com.joyent.manta.http.HttpHelper MantaClient.httpHelper                    N/A
     24     4                                           java.lang.String MantaClient.home                          N/A
     28     4                      com.joyent.manta.config.ConfigContext MantaClient.config                        N/A
     32     4                                              java.util.Set MantaClient.danglingStreams               N/A
     36     4                                java.lang.ref.WeakReference MantaClient.signerRef                     N/A
     40     4                          com.joyent.manta.client.UriSigner MantaClient.uriSigner                     N/A
     44     4               com.joyent.manta.client.MantaMBeanSupervisor MantaClient.beanSupervisor                N/A
     48     4                      com.joyent.manta.config.ConfigContext LazyMantaClient.configTracked             N/A
     52     4         com.joyent.manta.client.AuthenticationConfigurator LazyMantaClient.auth                      N/A
     56     4   com.joyent.manta.http.MantaConnectionFactoryConfigurator LazyMantaClient.connectionConfig          N/A
     60     4               com.joyent.manta.client.MantaMBeanSupervisor LazyMantaClient.lazyBeanSupervisor        N/A
     64     4                                          java.util.EnumSet LazyMantaClient.reload                    N/A
     68     4                           com.joyent.manta.http.HttpHelper LazyMantaClient.lazyHttpHelper            N/A
     72     4                          com.joyent.manta.client.UriSigner LazyMantaClient.lazyUriSigner             N/A
     76     4                                                            (loss due to the next object alignment)
Instance size: 80 bytes
Space losses: 3 bytes internal + 4 bytes external = 7 bytes total
```

Heavily considering the move to an abstract base class, it would be less weird than this implementation but it would mean having 3 classes in the `MantaClient` hierarchy instead of just 2. That doesn't seem so bad right?

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/LazyMantaClient.java
```diff
@@ -0,0 +1,299 @@
...
+        return new StandardHttpHelper(connectionContext, getContext());
+    }
+
+    // LIFECYCLE METHODS
+
+    /**
+     * Closed clients have cleaned up resources and should no longer be used.
+     */
+    private void ensureClientNotClosed() {
+        if (isClosed()) {
```
dekobon: Please give the user of the SDK more information so they know what they did wrong and/or what were the circumstances that led them to this exception.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/LazyMantaClient.java
```diff
@@ -0,0 +1,299 @@
...
+     * components. If provided the {@link MantaConnectionFactoryConfigurator} will be passed on to a
+     * {@link MantaConnectionFactory} when the {@link HttpHelper} is being rebuilt.
+     *
+     * @param config           the {@link ConfigContext} to track
+     * @param connectionConfig customized connection objects to use when creating the {@link HttpHelper}, or null
+     */
+    public LazyMantaClient(final ConfigContext config,
+                           final MantaConnectionFactoryConfigurator connectionConfig) {
+        super();
+        this.configTracked = config;
```
dekobon: You may want to consider a null guard clause here so that the contract for non-null values is explicit.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1690,16 +1739,6 @@ private void moveDirectory(final String source, final String destination,
         deleteRecursive(source);
     }
 
-    /**
```
dekobon: Removing this will break a lot of libraries that need to do funky things to introspect the state of a `MantaClient`.

dekobon: Hmnn.. Maybe this just got moved. Can you double check?

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaLazyConfigurationException.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.joyent.manta.exception.MantaException;
+
+/**
+ * Exception for errors occurring as a result of dynamic client reconfiguration.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
+ * @since 3.1.7
+ */
+public class MantaLazyConfigurationException extends MantaException {
```
dekobon: I'm curious why this inherits from `MantaException` and not `ConfigurationException`.

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/LazyMantaClientTest.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /**
+     * Verify that interceptors are not carried over between reloads. Looking at the implementation of
+     * {@link org.apache.http.protocol.HttpProcessorBuilder}'s usage of a deduplicating
+     * {@link org.apache.http.protocol.ChainBuilder} means that the shallow clone in
+     * {@link MantaConnectionFactoryConfigurator} is not necessary for the correctness of the
+     * {@link CloseableHttpClient} that is ultimately built. On the other hand, every time a configuration
+     * change occurs, another interceptor would be added to the provided {@link HttpClientBuilder} and
+     * the memory usage of that builder would grown indefinitely. See {@link canReloadHttpHelperRepeatedlyWithoutLeakingInterceptors()}
```
dekobon: The link isn't defined properly. Be sure to run javadoc and see if there are any errors.

java-manta-client-unshaded/src/test/java/com/joyent/manta/util/HttpClientBuilderShallowClonerTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.util.LinkedList;
+
+import static org.mockito.Mockito.mock;
+
+@Test
+public class HttpClientBuilderShallowClonerTest {
```
dekobon: You may want to write a test or just try using the clone with a `SecurityManager` that denies reflection and see if the exception message output is useful when we get a report of it not working.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.joyent.manta.client;
+
+import com.joyent.manta.config.ConfigContext;
+
+/**
+ * TODO: this might belong in the main package?
+ */
```
dekobon: Let's eject the TODO and put an issue in the issue tracker.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/AuthenticationConfigurator.java
```diff
@@ -0,0 +1,165 @@
...
+    public AuthenticationConfigurator(final ConfigContext config) {
+        this.config = config;
+        reload();
+    }
+
+    /**
+     * Check the configuration for authentication-related changes.
+     */
+    public void reload() {
+        synchronized (configureLock) {
```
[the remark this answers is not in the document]
tjcelaya: Cargo-culting the "keep your locks private" advice, it seems unnecessary now that I realize we don't expose the `AuthenticationConfigurator` to users. I'll switch to the simpler `synchronized` method usage.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/LazyMantaClient.java
```diff
@@ -0,0 +1,299 @@
...
+import com.joyent.manta.http.MantaConnectionFactory;
+import com.joyent.manta.http.MantaConnectionFactoryConfigurator;
+import com.joyent.manta.http.StandardHttpHelper;
+import org.apache.commons.lang3.BooleanUtils;
+
+import java.util.EnumSet;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * A lazily-initializing MantaClient. Users are expected to pass a
```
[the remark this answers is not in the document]
tjcelaya: I started out with this goal in mind but dialed it back as things got more complicated. I'll need to revisit locking of the `EnumSet<MantaClientComponent>` field and put together some  test cases using `CountDownLatch` before I'd feel comfortable declaring this class thread-safe. If that proves too difficult I'm assuming we'd want to add `@NotThreadSafe` (at least until the thread safety is added).

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
tjcelaya: Moved alongside the other getters.

tjcelaya:
@dekobon @cburroughs Added documentation and an example, ready for a final pass. The above summary comment has also been updated and now refers to the current implementation.

tjcelaya:
Ok actually ready for final review. The `AuthenticationConfigurator` class has been moved to the config package and renamed to `AuthAwareConfigContext`.

USAGE.md
```diff
@@ -344,3 +346,15 @@ Users leveraging this feature should be comfortable with the internals of the Ap
 familiarity with the
 [`MantaConnectionFactory`](/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java)
 class is recommended.
+
+### Modifying configuration after client instantiation
+
+Some configuration parameters can be updated in the `ConfigContext` after the client has already been constructed.
+Instead of directly providing a `ConfigContext` users can instead supply a `AuthAwareConfigContext` to the `MantaClient`
+constructor which can be programmatically updated using its `reload()` method. Because the `MantaClient` can be used in a
+multi-threaded fashion, users should be careful to await termination of in-flight requests in before triggering authentication reload.
```
cburroughs: If users need to stop all requests, call `reload()`, and then start up again, when would I choose this over just creating a new `MantaClient`?

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/AuthAwareConfigContext.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Object for tracking configuration changes related to authentication and recreating dependent objects as needed.
+ * It combines the users' configuration with the derived runtime objects needed to authenticate requests.
+ * objects like the {@link ThreadLocalSigner} which needs careful lifecycle management.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
+ * @since 3.1.7
+ */
+public class AuthAwareConfigContext
```
cburroughs: This class appears to have inconsistent synchronization.  For example, writes to `keyPair` are `synchronized` but reads or not, so a threads may see the updated fields at different times (or never!) and different orders.

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/KeyPairFactory.java
```diff
@@ -53,17 +53,18 @@ public KeyPair createKeyPair() {
         final String keyPath = config.getMantaKeyPath();
         final String password = config.getPassword();
 
+        final char[] charPassword;
+        if (password != null) {
+            charPassword = password.toCharArray();
+        } else {
+            charPassword = null;
+        }
+
```
cburroughs: Is this just refactoring this block out of the `try`?

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/DynamicHttpSignatureRequestInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+            return;
+        }
+
+        final long start = System.nanoTime();
+        final HttpSignatureAuthScheme authScheme = authConfig.getAuthScheme();
+        final Header authorization = authScheme.authenticate(authConfig.getCredentials(), request, context);
+        final long end = System.nanoTime();
+
+        request.setHeader(authorization);
+        request.setHeader("x-http-signing-time-ns", String.valueOf(end - start));
```
cburroughs: Is this header new? How does it relate to lazy config?

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/UriSigner.java
```diff
@@ -95,10 +98,11 @@ public URI signURI(final URI uri, final String method, final long expires)
 
 
         StringBuilder request = new StringBuilder();
-        final byte[] sigBytes = sigText.toString().getBytes(
-                StandardCharsets.UTF_8);
-        final byte[] signed = signer.get().sign(config.getMantaUser(),
-                                                keyPair, sigBytes);
+        final byte[] sigBytes = sigText.toString().getBytes(StandardCharsets.UTF_8);
+
+        // first parameter isn't actually used for anything, just checked for nullness
```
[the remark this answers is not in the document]
tjcelaya: You seem to have linked to `Signer#createAuthorizationHeader(String, KeyPair, String)` but the method being called here is [`Signer#sign(String, KeyPair, byte[])`](https://github.com/joyent/java-http-signature/blob/master/common/src/main/java/com/joyent/http/signature/Signer.java#L370).

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -295,6 +254,26 @@ private static void dumpConfig(final ConfigContext context) {
         }
     }
 
+    /**
+      * Method that returns the configuration context used to
+      * instantiate the MantaClient instance.
+      *
+      * @return instance of the configuration context used to construct this client
```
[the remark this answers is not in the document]
tjcelaya: This needs to be updated now that all `ConfigContext`s which are not `AuthAware` get converted during the constructor. I'll update the wording.

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/KeyPairFactory.java
tjcelaya: The password can now be supplied for both key content and key paths. Passwords were only checked in the `privateKeyContent != null` branch. Notice the usage on [this line](https://github.com/joyent/java-manta/pull/357/files/ed0f307ce37f9fbeebf7c2df94ddce2be67b598f#diff-273d69ef829954db5a617c6424225dd3R67).

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/DynamicHttpSignatureRequestInterceptor.java
tjcelaya: This class reimplements 
[HttpSignatureRequestInterceptor](https://github.com/joyent/java-http-signature/blob/master/apache-http-client/src/main/java/com/joyent/http/signature/apache/httpclient/HttpSignatureRequestInterceptor.java) using the `AuthAwareConfigContext`.

USAGE.md
tjcelaya: Attempting to provide flexibility for interactive applications. In the case of Cyberduck the client objects are created before all the authentication parameters are available.

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/AuthAwareConfigContext.java
tjcelaya: Created an inner class that holds the derived objects and sits inside an `AtomicReference`. Please review the new behavior.

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/AuthAwareConfigContext.java
```diff
@@ -0,0 +1,227 @@
...
+        reload();
+    }
+
+    /**
+     * Check the configuration for authentication-related changes. Clean up old authentication objects which might
+     * still exist and load new instances.
+     *
+     * @return this after reload
+     */
+    public synchronized AuthAwareConfigContext reload() {
```
cburroughs: I've been staring a this `synchronized` with doubt as to its necessity and I'm afraid I have another concern.  If thread A calls `setEncryptionPrivateKeyPath`, thread B calls `setMantaKeyId`, and thread C calls `reload()` I *think* we have a potential problem since this `synchronized` is effectively protecting the read of all fields, but not the writes.  I don't have my copy of JCIP on hand to verify with the right term.  Please let me know if that is confusing.

One possibility to avoid a lot of what would usually be unnecessary synchronization would be to create a ConfigContext that uses a `ReentrantReadWriteLock` and then inherent from that.

tjcelaya: You're right, the modification of the bag of parameter values in the config context is not totally threadsafe. I didn't override all the setters in an effort to keep the PR easy to review, and while I acknowledge that concurrent writes to the `ConfigContext` are a possibility I wanted to avoid complicating the general usage to accommodate an edge case within an edge case. I'll experiment with a `ReentrantReadWriteLock` and see how much that complicates `AuthAwareConfigContext`.

tjcelaya: After further consideration don't really want to go as far as wrapping all the config values in a lock since some of them are accessed extremely often. Would documenting the thread-unsafety of concurrent modifications to `AuthAwareConfigContext` be adequate? Even with locking there's no way to guarantee that a single consistent version of a config will be used for the entire request since the client may read from the `ConfigContext` anytime while executing a request (e.g. verifyUploads is checked late in the request cycle relative to username)

dekobon: One thought that might be out there - you could have all of the setters and getters wrap to operations that store their data in a `ConcurrentHashMap` instance. By doing so, you should get an optimistic locking implementation by default. In fact, you may be able to inherit from `MapConfigContext` and just supply it with the `ConcurrentHashMap`.

tjcelaya: The issue @cburroughs is getting at is more about synchronizing the creation of derived objects with updates to the underlying configuration itself. Even if the parameters themselves were allowed to be accessed concurrently we'd still need a lock that around parameters being updated while we're inside the `reload` method.

tjcelaya:
Does this comment include your review of the latest changes? Locking for the setters and reload does make the class longer but minimizes the complexity for maintainers. I'm wary of the mental burden of bringing in a `ReentrantReadWriteLock` in order to partially solve the transactional-multi-field-update problem. Given the existing behavior and usage patterns of the `ConfigContext` classes that design goal seems like scope creep, though it would still fall short of being a fully transactional `ConcurrentConfigContext` which manages read and write locks around *all* configuration state. I agree that synchronizing updates to the config with respect to setters and derived fields is within the scope of this change so I've limited the locking and `volatile` usage to address your concerns but externalizing the lock so it can be used across multiple fields (or offering some kind of begin-commit-rollback functionality) sounds like an entirely new problem to me. It seems like we're bumping into #315 (i.e. a SettableConfigContext is not always the source of truth for what the client uses to make requests, which this PR partially touches on) so I'm wary of tossing more features into this already-too-large PR before we can agree on new behavior in that issue.

tjcelaya:
I don't want it to seem like I am against the idea of a `ConcurrentConfigContext` for multi-field updates but it seems more like a feature request than an actual issue with this design.

As an example, `ConcurrentHashMap` is not a `HashMap` that offers multi-key synchronization, it's a `HashMap` that allows multi-threaded *access* to separate keys within the map. Thread-safety and transactions are slightly different topics.

cburroughs:
In attempting to pull apart the cases I'm afraid I left the discussion muddy.  I agree that we don't need to solve the "transaction" case (B) here, and I'm not even sure it is a real problem that is worth the cost of solving. My primary concern is the correctness case (A).

Locking on setters and `reload()` only is insufficient even in the simple #311 motivating case:
 * Thread A creates a MantaClient
 * Thread B calls whatever resources resources are needed for configuration, calls the lock protected setters, and calls reload().
 * Time passes.
 * Thread C calls `getAllJobIds` which calls `config.getMantaURL()`.  Because this *read* is not synchronized in the same way as the write, we may not see the value written by thread B.

Given the constraint of correctness, I agree that "complexity for maintainers" is the right thing to minimize.  What I hope to be simple when I describe in prose may not be in practice, or simplicity is not always an objective thing.   For me, any case where we are only *sometimes* locking access to a field is time consuming to reason about.  It feels to me like one of the most complex things we could ask a maintainer to keep working.

A class of bean-esque class of getters and setters where each method uses `synchronized` I would consider relatively simple because it is `@ThreadSafe` with no caveats.  The same class with a RWLock is unfortunately much more boilerplate, but I think also straightforward to reason about because it is `@ThreadSafe`.   That is what I meant to suggest by `ConcurrentConfigContext`.

I also suggested that we could expose that lock to subclasses for `reload()` to mitigate the "transaction" problem.  I admit this was speculative and presumed that a minor mitigation to the "transaction" problem being worth while.  I think you are right that the benefits to users for such a partial mitigation are very small, while for maintainers it would no longer quality as "boring" java concurrency.

cburroughs:
What is the plan for resolving the problem outlined [above](https://github.com/joyent/java-manta/pull/357#issuecomment-339479599).  The current locking is insufficient even in the simple #311 motivating case.

tjcelaya:
We could make all the standard config fields `volatile`. I'm hesitant to put a read lock in front of every config property access but that should address your concern about multi-threaded visibility of individual properties.

cburroughs:
Given your preference and the motivating case I think `volatile` sounds like a good path forward.

PR #364, opened 2017-10-23
Remove redundant integration test runs

### LINKED ISSUE #336: testng encryptionCipher parameter does not appear to do anything
Running with a scaled down testng xml file:

```
    <test name="Manta Client Integration Tests [AES128/CBC Encrypted]">
        <parameter name="usingEncryption" value="true"/>
        <parameter name="encryptionCipher" value="AES128/CBC/PKCS5Padding"/>
        <packages>
            <package name="com.joyent.manta.client.*" />
        </packages>
    </test>
```

In `IntegrationTestConfigContext.enableTestEncryption` from simple `println` additions, the value of `encryptionCipher`  is always `null` and `cipherDetails.getCipher().getAlgorithm()` is `AES/CTR/NoPadding`.  Looking at the class more broadly, there is code to check system properties and environmental variables, but I don't see anything invoking the testng parameter system.

In light of #155 #154 #202 #201 I'm not sure what the expected status of anything other than the default is.

tjcelaya:
# Overview

Resolves #336 and adds a `it.dryrun` property that can be used to list the tests that would run with their respective TestNG params. Unfortunately it looks like TestNG doesn't completely support filtering a test suite by groups so we need to be more explicit about suite definition.

Output of `it.dryrun` using current `testng-it.xml` (with some fixes to allow it to be used with `TestListingInterceptor`:
```
DRY-RUN: Listing [603] tests that would have run
```
Count of test cases that would run using new `testng-it.xml`:
```
DRY-RUN: Listing [405] tests that would have run
```

Lists for both are [here](https://gist.github.com/tjcelaya/fa82e54bd4be55ec95e1146954d7bd1f). Note that the `dry-run-after` list is slightly inaccurate since it shows different values for `encryptionCipher`. Other changes in this branch include using `encryptionCipher` so a dry-run listing of `master` would show `AES128/CBC/PKCS5Padding` for all non-null ciphers in that list.

# Changes

- change `usingEncryption` (boolean) to just `encryptionCipher` (string). omission (null) disables encryption
- add a new constructor for `IntegrationTestConfigContext` which handles the above
- switch from `@BeforeClass` with `@Parameters` to test constructors with just `@Parameters`. Whenever `SkipException` was thrown or `putDirectory` was called in a test which needed `@Parameters` a minimal method was created for `@BeforeClass`. The goal here was to make it safe to instantiate test classes without incurring side-effects.
- remove encryption param from tests which don't actually perform any object PUTs or MPU uploads (e.g. `MantaClientDirectoriesIT`)
- change `<packages>` to `<classes>` and target specific classes. I was going to use test groups for this but it seems like TestNG has trouble with mixing group filters with packages/classes (e.g. the comment in `testng-it.xml` about how we can't use `<groups><run><exclude>multipart</exclude></run></group>`
- document a new system property `it.dryrun` which can be used by the new interceptor. When this system property is truthy as defined by `BooleanUtils#toBooleanObject`, i.e. the following are all true: `y`, `Y`, `t`, `T`, `on`, `ON`, `yes`, `YES`, `true`, `TRUE` and all the mixed-case versions thereof
- add a test listener called `TestListingInterceptor` which can be used to collect suite definitions and disable actually running tests in the event `it.dryrun` is truthy


cburroughs:
Reminder:  On the last run integration tests had 64% instruction coverage and 39% branch coverage.  After this lands those numbers should stay the same.

tjcelaya:
For the dry-run aspect of this PR see https://github.com/cbeust/testng/issues/1503. Updating to TestNG version 6.14 and using the `-Dtestng.mode.dryrun=true` option can be used to perform the test listing.

nairashwin952013:
This PR has been re-introduced as a different PR #504 since the changes introduced in #364 cannot be holistically integrated into the current SDK code.

java-manta-it/src/test/java/com/joyent/manta/client/multipart/EncryptedServerSideMultipartManagerSerializationIT.java
```diff
@@ -52,24 +53,26 @@
 
     private ConfigContext config;
 
-    @BeforeClass()
-    @Parameters({"usingEncryption"})
-    public void beforeClass(@org.testng.annotations.Optional Boolean usingEncryption) throws IOException {
+    @Parameters({"encryptionCipher"})
+    public EncryptedServerSideMultipartManagerSerializationIT(final @Optional String encryptionCipher) throws IOException {
```
nairashwin952013: A ```compilation-error``` will be generated here since this ```integration-test``` will mandatorily require invocation of the TestNG parameter ```encryptionCipher```

PR #372, opened 2017-11-01
Allow users to optimize `putDirectory` with new tunable `manta.skip_directory_depth`

### LINKED ISSUE #371: Attempt to skip redundant `putDirectory` calls if possible
As a performance optimization we'd like to avoid `PUT` on directories that already exist.

tjcelaya:
Resolves #371. Due to the performance impact of redundant PUT requests on the server we want to backport this fix onto 3.1.6 and test its performance in isolation from the other changes that have occurred on master in the interim. I'll take care of applying this change to master later.


tjcelaya:
It actually makes more sense to work from the bottom up, i.e. check the deepest directory, check the parent of the deepest, check that parent's parent, etc. until you hit the deepest directory that _does_ exist and then proceed downward. I'll compare the performance of these for up to 10 levels of testing before deciding.

dekobon:
If you want to optimize the operation, it may be best to not check and skip the HEAD request entirely. Then you can just handle the error returned such that you use it to fork the logic to create the missing resources.

tjcelaya:
@dekobon @cburroughs Given the possibility that the probing algorithm described above _could_ result in more calls than the pessimistic case, should we start looking into strategies related to caching or would we want a more comprehensive time-complexity analysis before abandoning this approach?

tjcelaya:
Discussed some options for this with @dekobon yesterday and am now considering expanding the strategy pattern to allow users to implement an interface that accepts a path and can indicate "you can skip these parts of that path" based on the user's access pattern. This would give users a large degree of flexibility in optimizing their directory `PUT`s at the expense of requiring external code to achieve this optimization.

A _much_ simpler option would be to introduce the concept of a "working directory," i.e. `MANTA_WORKING_DIRECTORY = "/user/stor/foo/bar"`. The client could check that directory for existence when starting up (and potentially create it if it's missing). This would allow us to save _obviously_ redundant `PUT` calls at the expense of and extra `String` (or perhaps a `String[]` for the prefixes) worth of state in the client and avoids attempting to design one-fits-all caching or probing schemes. Additionally it would be much simpler for users to take advantage of than the above `RecursiveDirectoryCreationSkippingStrategy` described in the previous paragraph.

tjcelaya:
Noting that the `MANTA_WORKING_DIRECTORY` idea does not map well to services which handle many users. A single path would not cover data{base,warehouse,lake} use-cases where operations are rooted at a certain depth and contain a fixed set of child directories which are stable and only vary past a certain depth.

cburroughs:
> expanding the strategy pattern to allow users to implement an interface

I think this sounds like the best path forward for the reasons you described.  java-manta could still include a set of strategies that we maintain, but we are not trying to make a magical one size fits all.

> should we start looking into strategies related to caching

For the case of creating creating deeply nested directories and never deleting them, a bloom filter might be a natural fit.

tjcelaya:
We'll probably need to limit the number of directories that go in the bloom filter in order to minimize (or at least bound) false positives. Memory usage and concurrent access would become issues though.

dekobon:
I was thinking that a hash compaction approach may work better because we already know the size of all possible values. @cburroughs thoughts?

tjcelaya:
Can we keep in mind the scope of functionality being discussed? We're trying to minimize redundant `PUT`s for users which have noted they're spending time repeatedly creating directories as a result of defensive calls to `putDirectory`.

When the use case is migration from S3 and directories are a stumbling block, asking users to implement an interface _in addition to_ all the client configuration we already require doesn't seem very friendly to me. Frankly I was afraid that leaving `RecursiveDirectoryCreationStrategy` in the PR would skew the conversation towards more esoteric solutions when it was intended solely to ease exploration.

The bloom filter (family of) suggestion(s) _could be_ tractable if we wanted to spend time to benchmark different implementations, validate thread safety, and found a straightforward way to describe tuning, but the questions are endless:

 - Do users want to tune the redundant directory cache based on depth, number of directories, branching factor?
  - What if users *do* occasionally delete directories, does that mean we discard the bloom filter entirely or follow the rabbit hole deeper towards more advanced solutions like [Invertible Bloom Filter](https://arxiv.org/pdf/1101.2245.pdf) and [Cuckoo Filter](https://github.com/seiflotfy/cuckoofilter)?
  - Run time of an individual client becomes a concern for some of these approaches and would need to be factored into tuning the configuration of such data structures. [Scalable bloom filters](http://gsd.di.uminho.pt/members/cbm/ps/dbloom.pdf) anyone?

At the end of the day, we'd still be talking about "how much memory and overhead do we want to throw at skipping the creation of existing directories."

As attractive a solution as [hash compaction](http://verify.stanford.edu/PAPERS/SD96B.pdf) might seem it puts us even further into unknown territory. I cringe at the thought of a future maintainer being asked to look into "an issue with `putDirectory` occasionally failing" and stumbling into what is effectively a hand-rolled bloom filter that _might_ have a bug or might just be saturated and now reports that **every** directory exists because the client was running for several days.

java-manta-client/src/main/java/com/joyent/manta/config/SettableConfigContext.java
```diff
@@ -291,6 +299,14 @@ static void setAttributeFromContext(final String name,
             case EnvVarConfigContext.MANTA_TCP_SOCKET_TIMEOUT_ENV_KEY:
                 config.setTcpSocketTimeout(MantaUtils.parseIntegerOrNull(value));
                 break;
+            case MapConfigContext.MANTA_UPLOAD_BUFFER_SIZE_KEY:
+            case EnvVarConfigContext.MANTA_UPLOAD_BUFFER_SIZE_ENV_KEY:
+                config.setUploadBufferSize(MantaUtils.parseIntegerOrNull(value));
```
tjcelaya: This option was missing from this `switch` but is unrelated to `MANTA_SKIP_DIRECTORY_DEPTH`

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1431,24 +1431,20 @@ public void putDirectory(final String rawPath, final boolean recursive,
...
-
-        for (int i = 0; itr.hasNext(); i++) {
-            final String part = itr.next().toString();
-            sb.append(part);
+        final RecursiveDirectoryCreationStrategy directoryCreationStrategy;
+        if (config.getSkipDirectoryDepth() != null) {
+            directoryCreationStrategy = new DepthSkippingDirectoryCreationStrategy(this, config.getSkipDirectoryDepth());
+        } else {
+            directoryCreationStrategy = new PessimisticDirectoryCreationStrategy(this);
+        }
```
tjcelaya: We select a `directoryCreationStrategy` each time `putDirectory` is called so that the configuration parameter can be meaningfully tuned from the `ConfigContextMBean` in JMX.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1431,24 +1431,20 @@ public void putDirectory(final String rawPath, final boolean recursive,
...
-                putDirectory(sb.toString(), headers);
-            }
+        putDirectory(rawPath, headers, directoryCreationStrategy);
+    }
 
-            if (itr.hasNext()) {
-                sb.append(SEPARATOR);
-            }
-        }
+    void putDirectory(final String rawPath,
```
tjcelaya: Leaving this method package-private for testing only.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1431,24 +1431,21 @@ public void putDirectory(final String rawPath, final boolean recursive,
...
 
-        final String[] parts = rawPath.split(SEPARATOR);
-        final Iterator<Path> itr = Paths.get("", parts).iterator();
-        final StringBuilder sb = new StringBuilder(SEPARATOR);
-
-        for (int i = 0; itr.hasNext(); i++) {
-            final String part = itr.next().toString();
-            sb.append(part);
+        final Integer skipDepth = config.getSkipDirectoryDepth();
+        final RecursiveDirectoryCreationStrategy directoryCreationStrategy;
```
tjcelaya: We select a directoryCreationStrategy each time putDirectory is called so that the configuration parameter can be meaningfully tuned from the ConfigContextMBean in JMX.

(duplicated for clarity, was collapsed by a pushed commit)

tjcelaya:
Added some review comments to clarify a few things.

java-manta-client/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Base class for classes which implement recursive directory creation.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
+ * @since 3.1.7
+ */
+abstract class RecursiveDirectoryCreationStrategy {
+
+    @SuppressWarnings("checkstyle:JavaDocVariable")
```
dekobon: Let's see if we can globally suppress this check in our checkstyle settings because this is just silly.

java-manta-client/src/main/java/com/joyent/manta/client/DepthSkippingDirectoryCreationStrategy.java
```diff
@@ -0,0 +1,95 @@
...
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("ASSUME " + assumedExistingDirectory);
+        }
+
+        final Boolean redundantPut = createNewDirectory(maybeNewDirectory, headers);
+        incrementOperations();
+
+        if (redundantPut == null) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("FAILED:" + maybeNewDirectory);
```
dekobon: Nitty nit nit nitpick: You can avoid doing the `isXXEnabled()` check if you do:
```java
LOG.debug("FAILED: {}", maybeNewDirectory);
```

java-manta-client/src/main/java/com/joyent/manta/client/DepthSkippingDirectoryCreationStrategy.java
```diff
@@ -0,0 +1,95 @@
...
+        if (paths.length <= skipDepth) {
+            createPessimistically(rawPath, headers);
+            return;
+        }
+
+        final String assumedExistingDirectory = paths[skipDepth - 1];
+        final String maybeNewDirectory = paths[skipDepth];
+
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("ASSUME " + assumedExistingDirectory);
```
dekobon: Same comment as the other debug statements.

java-manta-client/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * The client to use when performing operations.
+     */
+    private final MantaClient client;
+
+    /**
+     * The number of PUT operations this instance has performed.
+     */
+    private int operations;
```
dekobon: Consider `java.util.concurrent.atomic.LongAccumulator` for counting the number of operations.

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -1431,24 +1431,21 @@ public void putDirectory(final String rawPath, final boolean recursive,
...
+        if (skipDepth != null && 0 < skipDepth) {
+            directoryCreationStrategy = new DepthSkippingDirectoryCreationStrategy(this, skipDepth);
```
dekobon: Is there a way we author the strategy such that we don't need to create a new object per put operation?

java-manta-client/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return whether or not the directory was actually new, or null if it failed to be created
+     */
+    Boolean createNewDirectory(final String path, final MantaHttpHeaders headers) throws IOException {
+        try {
+            return client.putDirectory(path, headers);
+        } catch (final MantaClientHttpResponseException mchre) {
+            if (mchre.getServerCode().equals(MantaErrorCode.DIRECTORY_DOES_NOT_EXIST_ERROR)) {
+                return null;
+            } else {
+                throw mchre;
```
dekobon: Is there any additional error detail that we want to capture here?

java-manta-client/src/main/java/com/joyent/manta/util/MantaUtils.java
```diff
@@ -373,6 +376,44 @@ public static String lastItemInPath(final String path) {
...
     }
 
+    /**
+     * Compute the intermediate directories between "/" and {@code rawPath}.
+     *
+     * @param rawPath The fully qualified path of the Manta directory.
+     * @return All non-root directories leading to and including rawPath
+     */
+    public static String[] prefixPaths(final String rawPath) {
+        final String[] parts = rawPath.split(SEPARATOR);
```
dekobon: We may want to skip using `String.split()` here and use `StringUtils.split()` because we don't need regular expression parsing. However, things have changed enough in the JDK, that this may be just superstition and old habits on my part.

EDIT: I'm not the only one confused: https://stackoverflow.com/a/11002374/33611

java-manta-client/src/main/java/com/joyent/manta/client/MantaClient.java
tjcelaya: If the extra object per recursive-`PUT` is a serious concern we could convert the strategies to static methods that use their return value to convey the number of operations taken. This would leave us with just the `RecursiveDirectoryCreationStrategy` class having no constructor and two methods:

`int complete(MantaClient, String, MantaHttpHeaders)`
`int skippingDepth(MantaClient, String, MantaHttpHeaders, int)`

java-manta-client/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java
tjcelaya:
```<module name="JavadocVariable">
    <property name="ignoreNamePattern" value="LOG|LOGGER"/>
</module>```

added to `checkstyle.xml`

File: java-manta-client/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java (hunk 10 above)
Comment (tjcelaya) [reply java-manta-client/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java]: The stack trace should indicate that we were in the middle of a recursive directory creation but we could also add one or both of these to the context:

```
recursiveDirectorySkipStart = $FAILED_PUT
recursiveDirectorySkipTarget = $FINAL_CHILD_PATH
```

java-manta-client/src/main/java/com/joyent/manta/util/MantaUtils.java
tjcelaya: Since `SEPARATOR` is a string I ended up with `StringUtils.split(rawPath, SEPARATOR.charAt(0))`. The output is different as well, previously I'd get an empty string as the first element since the string started with `/`. One less unneeded loop.

PR #380, opened 2017-11-23
Sign encoded URLs

### LINKED ISSUE #379: getAsSignedURL does not handle URL encoding properly
Reported by a Mountain Duck user. Any path with a space in it (or other character that needs to be URL-encoded) will sign just fine but result in:

```
{
"code": "InvalidSignature",
"message": "The signature we calculated does not match the one you sent"
}
```

tjcelaya:
Resolves #379 and https://trac.cyberduck.io/ticket/10151

We should be using the encoded URL for signing, e.g.:
```
/java.manta.ci/stor/%E2%9B%B0%20quack%20%F0%9F%A6%86
```
instead of
```
/java.manta.ci/stor/⛰ quack 🦆
```

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/UriSigner.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/UriSigner.java
@@ -72,10 +72,8 @@ public class UriSigner {
         Validate.notNull(method, "Method must not be null");
         Validate.notNull(uri, "URI must not be null");
 
-        if (uri.getQuery() != null) {
-            Validate.isTrue(StringUtils.isEmpty(uri.getQuery()),
-                    "Query must be null or empty. URI: %s", uri);
-        }
+        Validate.isTrue(StringUtils.isEmpty(uri.getQuery()),
+                "Query must be null or empty. URI: %s", uri);
 
         final ThreadLocalSigner signer = authConfig.getSigner();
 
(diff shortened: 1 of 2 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/client/UriSigner.java
```diff
@@ -72,10 +72,8 @@ public URI signURI(final URI uri, final String method, final long expires)
         Validate.notNull(method, "Method must not be null");
         Validate.notNull(uri, "URI must not be null");
 
-        if (uri.getQuery() != null) {
-            Validate.isTrue(StringUtils.isEmpty(uri.getQuery()),
-                    "Query must be null or empty. URI: %s", uri);
-        }
+        Validate.isTrue(StringUtils.isEmpty(uri.getQuery()),
```
tjcelaya: `StringUtils.isEmpty` returns true for `null` and empty strings

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -53,7 +56,7 @@ public void beforeClass() throws IOException {
 
         mantaClient = new MantaClient(config);
         testPathPrefix = IntegrationTestConfigContext.generateBasePath(config, this.getClass().getSimpleName());
-        mantaClient.putDirectory(testPathPrefix, null);
+        mantaClient.putDirectory(testPathPrefix, true);
```
tjcelaya: This test case can't be run on its own without changing this to recursive `putDirectory`

jussisallinen:
@tjcelaya Thanks a lot for the fix!

PR #397, opened 2018-01-06
Fix #395

### LINKED ISSUE #395: Reject non-ascii values in Manta Headers
Due to MANTA-3527, objects can go into a state where they cannot be read if the header was written with a non-ascii value. We should add a guard clause that tells users that a non-ascii value was invalid and guide them to encode it.

tjcelaya:
Need to figure out what's going on with the commit history below but the diff is correct.

Changes are restricted to `MantaMetadata`. Do we want to extend this fix to `MantaHttpHeaders` as well?

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaMetadata.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaMetadata.java
@@ -294,7 +294,7 @@ public class MantaMetadata implements Map<String, String>, Cloneable, Serializab
          */
         @Override
         public boolean evaluate(final String object) {
-            // NEXT: throw specific exceptions instead?
+            // NEXT: throw MantaIllegalMetadataException instead
 
             return object != null
                     && !object.isEmpty()
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaMetadata.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaMetadata.java
@@ -43,19 +43,24 @@ public class MantaMetadata implements Map<String, String>, Cloneable, Serializab
      * Prefix required for metadata keys being stored via HTTP headers on Manta.
      */
     public static final String METADATA_PREFIX = "m-";
+
     /**
      * Prefix required for encrypted metadata keys being stored in ciphertext.
      */
     public static final String ENCRYPTED_METADATA_PREFIX = "e-";
+
     /**
      * An array of characters considered to be illegal in metadata keys.
      */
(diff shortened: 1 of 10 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaMetadata.java
```diff
@@ -300,15 +336,15 @@ private boolean validPrefix(final String input) {
...
+        private boolean hasIllegalKeyChars(final String input) {
             final char[] chars = input.toCharArray();
 
             for (final char c : chars) {
                 if (isControlCharacter(c)) {
                     return true;
                 }
 
-                for (char illegalKeyChar : ILLEGAL_KEY_CHARS) {
+                for (final char illegalKeyChar : ILLEGAL_KEY_CHARS) {
```
dekobon: When writing this code, since this is in the hot path, I should have made the contents of `ILLEGAL_KEY_CHARS` sorted and used `Arrays.binarySearch()` to determine if a character is present in the values.

It might not matter and might be just a micro optimization.

java-manta-client-unshaded/src/test/java/com/joyent/manta/client/MantaMetadataTest.java
```diff
@@ -64,14 +68,37 @@ public void cantAddMetadataKeyThatIllegalCharacter() {
...
-            } catch (IllegalArgumentException e) {
+            } catch (final IllegalArgumentException e) {
                 caught = true;
             }
 
             Assert.assertTrue(caught, String.format("No exception thrown for char: %s", c));
         }
     }
 
+    // Until MANTA-3527 is resolved we should prevent users from using non-ascii values for both keys
```
dekobon: Even if MANTA-3527 is resolved, we will want to prevent the users from using non-ascii values in the headers / metadata.

tjcelaya: https://tools.ietf.org/html/rfc5987 seems to imply that ISO-8859-1 is the accepted character set but HTTP/1.1 suggests sticking to US-ASCII in the final paragraph of [this section](https://tools.ietf.org/html/rfc7230#section-3.2.4).

Since users can construct and manipulate their own `MantaHttpHeaders` objects does this mean we want to extend the validation to that class as well (i.e. use `HttpHeaderPredicate` in both classes)?

dekobon:
> Changes are restricted to `MantaMetadata`. Do we want to extend this fix to `MantaHttpHeaders` as well?

Yes, I think we will need to validate the encoding of the `MantaHttpHeaders` object as well.

Issue #410, opened 2018-04-09
Simplifying diagnosis of retries


tjcelaya:
It is currently impossible to determine how frequently the client is retrying requests without turning on DEBUG logging. Since there are many other logs at this logging level it would be helpful for users if we provided some warning or notice of the fact that we're having trouble performing requests.

Options we've identified include the following:
* Expose a retry rate in the existing `PoolStatsMBean` (or in a new bean). We can track this from `MantaHttpRequestRetryHandler`.
* Using the same retry-rate-tracking, log a higher-visibility message (INFO/WARN/ERROR instead of DEBUG) when the retry rate exceeds a certain threshold. We can set a sane default for this if we want to it to be enabled by default but we should be careful to not log this message too frequently.
* Embed an HTTP header in retry requests that can be tracked by the server-side as an indicator of "unhealthy" client activity.

We can do any or all or none of these options. we just need to be clear about why we select our option.

tjcelaya:
As described in the [final comment in #412](https://github.com/joyent/java-manta/pull/412#issuecomment-387142670), we'll need a total request count to provide context for the retry count. After some experimentation with the [HttpClient metrics](http://metrics.dropwizard.io/4.0.0/manual/httpclient.html) package, I've determined that we can't use that specific package since it relies on providing a custom implementation of `HttpRequestExecutor` and `HttpClientConnectionManager` classes from Apache HttpClient (which we already do ourselves). Thankfully, the [metrics-httpclient](https://github.com/dropwizard/metrics/tree/4.1-development/metrics-httpclient/src/main/java/com/codahale/metrics/httpclient) package is relatively small, so it should be straightforward to subclass `MantaHttpRequestExecutor` to add an instrumented version.

PR #412, opened 2018-04-15
Add retry metric

tjcelaya:
Notable changes:
- the package-private `MantaMBeanSupervisor` has been renamed to `MantaClientAgent` since it's now responsible for both MBeans and the metrics reporter. Relevant tests have been updated and expanded.
- Some private or package-private constructors now accept a `MetricRegistry` for either tracking their own metrics or passing it on to child objects. Constructor overrides have been added to supply `null` for previous usages of these constructors.
  - Current examples include:
    - `MantaHttpRequestRetryHandler` for tracking retries
    - `MantaClientAgent` for publishing the registry to JMX
- `IntegrationTestConfigContext` will have monitoring disabled by default (i.e. no MBeans and no metrics). It doesn't seem useful to care about JMX or metrics for integration tests right now. **This may be wholly or partially reverted**

This satisfies the first option in #410 and provides the necessary data for the second option. (note that neither of these exclude option 3) 

Other things to keep in mind:
 - Currently, a client generates its ID and passes that on to the agent. We may want to use this ID elsewhere (perhaps as part of the User-Agent?) and keep it in `MantaClient` or make this a local variable in the constructor. Either way, we need to address the `QUESTION` comment [near its initialization](https://github.com/joyent/java-manta/compare/master...tjcelaya:fix-410?expand=1#diff-a29b3a00d80d81587eee0c833d058a61R239).
 - We may want to mention the change in JMX structure in the changelog to reference client instance ID instead of "supervisor count." A screenshot of the new structure is included below. I believe the UUID for identifying clients disambiguates the relationship between `MantaMBeanSupervisor` and `MantaClient` for maintainers.

screen shot 2018-04-14 at 6 14 53 pm

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MetricReporterSupplier.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MetricReporterSupplier.java
@@ -31,10 +31,10 @@ final class MetricReporterSupplier implements Supplier<Closeable> {
      * The logger name to use for SLF4J reporting.
      */
     public static final String FMT_METRIC_LOGGER_NAME = "com.joyent.manta.client.metrics";
+
     /**
      * A metric reporter constructed and configured with settings from the supplied configuration.
      */
-
     private final Closeable reporter;
 
     /**
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MetricReporterSupplier.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MetricReporterSupplier.java
@@ -1,0 +1,86 @@
+package com.joyent.manta.client;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Reporter;
+import com.codahale.metrics.Slf4jReporter;
+import com.codahale.metrics.jmx.JmxReporter;
+import com.joyent.manta.config.MantaClientMetricConfiguration;
+import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
+
+import static org.apache.commons.lang3.Validate.inclusiveBetween;
+import static org.apache.commons.lang3.Validate.notNull;
+
+final class MetricReporterSupplier implements Supplier<Closeable> {
+
+    private final Closeable reporter;
+
+    MetricReporterSupplier(final MantaClientMetricConfiguration metricConfig) {
+        notNull(metricConfig);
+
+        switch (metricConfig.getReporterMode()) {
+            case JMX:
+                final JmxReporter jmxReporter = buildJmxReporter(metricConfig);
+                jmxReporter.start();
+                this.reporter = jmxReporter;
+                break;
+            case SLF4J:
+                final Slf4jReporter slf4jReporter = buildSlf4jReporter(metricConfig);
+                slf4jReporter.start(metricConfig.getPeriodicReporterOutputInterval(), TimeUnit.MILLISECONDS);
+                this.reporter = slf4jReporter;
+                break;
+            case DISABLED:
+            default:
+                this.reporter = null;
+        }
+    }
+
+    @Override
+    public Closeable get() {
+        return this.reporter;
+    }
+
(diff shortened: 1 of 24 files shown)


tjcelaya:
Pending changes:
- [ ] CHANGELOG

tjcelaya:
Additional user requirement (exposing metrics through logging instead of JMX) has been implemented. Still need to add the new options to USAGE.md, cleanup for checkstyle, and an update to the CHANGELOG.

tjcelaya:
Ready for another pass. In particular I'd like to know if the decision to use milliseconds as the time unit for `manta.metric_reporter.output_interval` makes sense.

I opted to use milliseconds for consistency with all the other time units makes sense. It might make more sense to use seconds, especially since [meters include 1-minute moving averages](http://metrics.dropwizard.io/4.0.0/manual/core.html#man-core-meters) and adding 1 extra log per second seems like overkill so millisecond precision would be even more ridiculous.

clstokes:
I have no comments for the changes since my last review. Looks good to me.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MetricReporterSupplier.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celaya</a>
+ */
+final class MetricReporterSupplier implements Supplier<Closeable> {
+
+    /**
+     * The logger name to use for SLF4J reporting.
+     */
+    public static final String FMT_METRIC_LOGGER_NAME = "com.joyent.manta.client.metrics";
+    /**
```
tjcelaya: misplaced newline

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
```diff
@@ -115,6 +115,16 @@
      */
     private Integer skipDirectoryDepth;
 
+    /**
+     * Whether metrics and MBeans should be tracked and exposed.
+     */
+    private volatile MetricReporterMode metricReporterMode;
+
+    /**
+     * Whether metrics and MBeans should be tracked and exposed.
```
tjcelaya: > metrics output interval in seconds for modes that report metrics periodically

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/MantaClientMetricConfiguration.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Value object describing how metrics should be exposed for a {@link com.joyent.manta.client.MantaClient}.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celaya</a>
+ * @since 3.1.9, 3.2.2
+ */
+public final class MantaClientMetricConfiguration {
+
+    /**
+     * Unique identifier for each {@link com.joyent.manta.client.MantaClient} instance. Used to avoid collisions when
```
tjcelaya: > Also used to avoid...

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java
```diff
@@ -352,7 +367,7 @@ protected HttpHost findProxyServer() {
             } else {
                 String msg = String.format(
                         "Expecting proxy to be instance of InetSocketAddress. "
```
tjcelaya: remove unnecessary newline and concat

java-manta-it/src/test/java/com/joyent/manta/config/IntegrationTestConfigContext.java
```diff
@@ -31,12 +31,14 @@
 
     private static String suiteRunId = UUID.randomUUID().toString();
 
+    private static final BaseChainedConfigContext BASE = new StandardConfigContext();
+
```
tjcelaya: undo since `enableTestEncryption` takes a `SettableConfigContext<T>` and would modify the `static BaseChainedConfigContext`, i.e. *this is bad*

tjcelaya:
After some time away from this PR I realized having a retry count without any idea of how many total requests are being sent doesn't seem very useful. This PR on its own is ready (and quite large) so I'll be opening a separate PR to implement at least the [per-method request count provided by `InstrumentedHttpClientConnManager` and `HttpClientMetricNameStrategies.METHOD_ONLY`](http://metrics.dropwizard.io/4.0.0/manual/httpclient.html).

PR #415, opened 2018-04-27
More helpful error messages for invalid keys

tjcelaya:
Bump dependency versions and provide error messages tailored to the way the key was provided. We need java-http-signature 4.0.8 because of [this issue](https://github.com/joyent/java-http-signature/issues/50) which causes an unhandled `NullPointerException`. The other dependency version changes are to satisfy dependency convergence. Release notes imply there are no breaking changes between versions.

Additionally, if the key was provided directly as key content the error message would say:

```
Unable to read key files from path: null
```

Now there are different error messages, one for a failure to load the key from a path:

```
Unable to read private key files from path: ${MANTA_KEY_PATH}
```

and a separate message for those providing the content:

```
Unable to read supplied private key content
```


tjcelaya:
This PR will fail until java-http-signature 4.0.8 is released.

PR #417, opened 2018-04-30
Resolve regression introduced by #357

tjcelaya:
Fixes `MantaClientRangeIT#canGetWithUnboundedStartRange` which was failing because of MANTA-3673 (inability to cope with anything but a single well-formed UUID in the `x-request-id` header).

~# Open Questions:~
I didn't think this through enough, fixing the test failure by updating `MantaHttpRequestFactory#addHeaders` seems like a quick fix for now and avoids the unexpected WARN log that users can't do anything about.

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaHttpRequestFactory.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaHttpRequestFactory.java
@@ -220,8 +220,8 @@ public class MantaHttpRequestFactory {
     }
 
     /**
-     * Add headers to an {@link org.apache.http.client.methods.HttpUriRequest} without clobbering defaults
-     * and authentication.
+     * Add headers to an {@link org.apache.http.client.methods.HttpUriRequest}. Never copies the request ID header.
+     * If reusing a request ID is desired (which it should never be) it will need to be done manually.
      *
      * @param httpMessage request to attach headers to
      * @param headers headers to attach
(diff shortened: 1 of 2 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaHttpRequestFactory.java
```diff
@@ -229,6 +229,10 @@ private void prepare(final HttpRequest request) {
     public static void addHeaders(final HttpMessage httpMessage, final Header... headers) {
         Validate.notNull(httpMessage, "HttpMessage must not be null");
         for (final Header header : headers) {
+            if (header.getName().equals(MantaHttpHeaders.REQUEST_ID)) {
```
dekobon: As a paranoia sufferer, I would invert the equals to avoid the remote possibility of a NPE:
```java
if (MantaHttpHeaders.REQUEST_ID.equals(header.getName())) {
```

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/RequestIdInterceptor.java
```diff
@@ -61,6 +61,12 @@
 
     @Override
     public void process(final HttpRequest request, final HttpContext context) throws HttpException, IOException {
+        if (request.containsHeader(MantaHttpHeaders.REQUEST_ID)) {
+            // See: MANTA-3673, manta barfs on malformed x-request-id
```
dekobon: I would log a warning and throw away the UUID and display the new UUID. Also, I would not check in the TODO.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaHttpRequestFactory.java
tjcelaya: Thanks, I was only taking existing calls to `addHeaders` into account.

tjcelaya:
I presume a failing build discourages reviewers from looking at a PR carefully, but I've been experimenting with calling out questions in the code vs. in the PR description. Apologies for the unsettling behavior.

It seems more straightforward to just avoid the misuse of `MantaHttpRequestFactory#addHeaders` altogether for now and avoid the risk of introducing logs that are not actionable for users.

PR #418, opened 2018-05-09
Per-HTTP-method request metrics and exception metrics

tjcelaya:
# Goal

We need to contextualize the "retries" metric described in #410 since knowing how many requests had  to be retried in the last minute isn't useful without knowing the number of requests completed within the last minute. While there is a convenience package in the form of [metrics-httpclient](http://metrics.dropwizard.io/4.0.0/manual/httpclient.html) it is implemented in a way that is not compatible with our current usage of `HttpClientBuilder`.

# Changes

Since `MantaHttpRequestExecutor` is a subclass of `HttpRequestExecutor`  and the `metrics-httpclient` package implements request metrics with its own subclass ([`InstrumentedHttpRequestExecutor`](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpRequestExecutor.java)), I've had to reimplement their solution in our own `InstrumentedMantaHttpRequestExecutor`. Fortunately, this means we can avoid pulling in the `metrics-httpclient` dependency which would've also forced us to revert our HttpClient version from 4.5.5 to 4.5.4.

# Open Questions

I'm most concerned about the usefulness of `InstrumentedPoolingHttpClientConnectionManager` which – unlike `PoolStatsMBean` – is compatible with both JMX and SLF4J reporting but doesn't really offer much of an improvement otherwise. I suspect it's also less performant since it registers 1 bean per gauge (for a total of 4) vs the singular MBean presented by `PoolStatsMBean`.

# Screenshots

screen shot 2018-05-08 at 5 33 24 pm

screen shot 2018-05-08 at 5 33 42 pm

screen shot 2018-05-08 at 5 34 07 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,12 @@ This project aims to adhere to [Semantic Versioning](http://semver.org/).
 ### Added
  - Client metrics can now be enabled by selecting a reporting mode with the
    [`manta.metric_reporter.mode`/`MANTA_METRIC_REPORTER_MODE`](/USAGE.md#parameters) setting. [JMX and SLF4J are
-   available, though others may be added in the future.](https://github.com/joyent/java-manta/issues/410#issuecomment-384751882)
- - [Retry rate metric](https://github.com/joyent/java-manta/issues/410).
+   available](https://github.com/joyent/java-manta/issues/410#issuecomment-384751882), though others may be added in the future.
+ - New metrics exposed:
+    - [Meter](http://metrics.dropwizard.io/4.0.0/manual/core.html#meter) for
+    [retries](https://github.com/joyent/java-manta/issues/410).
+    - [Timers](http://metrics.dropwizard.io/4.0.0/manual/core.html#timer) per HTTP request method.
+    - [Meters](http://metrics.dropwizard.io/4.0.0/manual/core.html#meter) per exception that occurs during requests.
 
 ### Changed
  - MBeans registered in JMX no longer use an incrementing integer and instead are created under
--- a/USAGE.md
+++ b/USAGE.md
@@ -168,9 +168,9 @@ Note: Dynamic Updates marked with an asterisk (*) are enabled by the `AuthAwareC
     (only `SLF4J` at present) requires also setting an output interval.
     See [the section on monitoring](#monitoring) for more information about reporting modes.
 * `manta.metric_reporter.output_interval` (**MANTA_METRIC_REPORTER_OUTPUT_INTERVAL**)
-    Integer interval in seconds at which metrics are reported by periodic reporters.
+    Nullable integer interval in seconds at which metrics are reported by periodic reporters.
     This number must be set and greater than zero if `manta.metric_reporter.mode`/`MANTA_METRIC_REPORTER_MODE`
-    is set to `SLF4J`.
+    is set to `SLF4J`. Defaults to `null`.
 * `manta.client_encryption` (**MANTA_CLIENT_ENCRYPTION**)
     Boolean indicating if client-side encryption is enabled.
 * `manta.encryption_key_id` (**MANTA_CLIENT_ENCRYPTION_KEY_ID**)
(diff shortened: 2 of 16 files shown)


dekobon:
I noticed that the new properties added to ConfigContext didn't appear in JMX:
image

dekobon:
Also, it is unclear what the default value for `manta.metric_reporter.output_interval` is when you are using JMX and you leave it unset.

dekobon:
Will metrics around socket timeouts or 500s be a different PR once this is integrated?

dekobon:
Additionally, do we want to capture the server side request time metric as well? I know we have that available in server-side metrics. However, can we imagine a case where the client wants that data?

I can imagine wanting to look at the server request latency graph at the same time as the total client latency graph. If I could compare the p99s of both side by side, it would allow me to rule out pathologies on the server.

tjcelaya:
> I noticed that the new properties added to ConfigContext didn't appear in JMX:

Added the missing properties to `ConfigContextMBean` in addition to reordering the fields in `BaseChainedConfigContext` to match the documentation (and calling out the intentionally-not-exported properties in the MBean to make validation easier).

> Also, it is unclear what the default value for `manta.metric_reporter.output_interval` is when you are using JMX and you leave it unset.

Clarified the documentation to mention that it is a _nullable_ integer and is left unset by default.

> Will metrics around socket timeouts or 500s be a different PR once this is integrated?

"Unexpected response code" is not handled by these metrics so I've opened #422 to track that. Could you elaborate on what kinds of metrics you'd expect regarding socket timeouts? In situations that trigger a `SocketTimeoutException` a meter would be added named "exceptions-SocketTimeoutException" to track the count and rate of occurrence (with the same properties as [the first screenshot](https://user-images.githubusercontent.com/1973223/39790026-400735fe-52e8-11e8-931d-644bef57c2db.png)).

> Additionally, do we want to capture the server side request time metric as well? ...

Updated #422 to include this as part of the investigation. Recording `x-response-time` by method (in the same was that `requests-$METHOD` is broken out) should be possible with a [histogram](https://metrics.dropwizard.io/4.0.0/manual/core.html#histograms) metric.

dekobon:
With regards to what kinds of metrics that I would expect regarding socket timeouts, I would expect the same type of metrics as shown in the screenshot for `NoHttpResponseException`. I would also additionally add the value of actual timeout setting. In theory, this value should be retrievable from the configuration context JMX properties. However, it is possible that it somehow got set via Java properties outside of our configuration mechanism, so it may be useful to show for that reason. Additionally, it placinging it in the same JMX property bucket also makes it very easy for users to understand what timeout is being exceeded without having to search for it.

tjcelaya:
Trying to figure out which timeout value was hit (generic "timeout" vs tcp socket "timeout") doesn't seem immediately obvious so we'll have to settle for the exception type. Frankly I get confused every time I try to trace the source of the request timeout without a debugger since we pass timeouts through both [`RequestConfig`](https://github.com/joyent/java-manta/blob/848f1f2cae72a416254750c035d1c40d36780bed/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java#L284) and the [connection manager](https://github.com/joyent/java-manta/blob/848f1f2cae72a416254750c035d1c40d36780bed/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaConnectionFactory.java#L254). After some digging it seems like [http.socket.timeout is deprecated](https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/params/CoreConnectionPNames.java) and is only referenced from other deprecated classes so I think it's actually no longer possible to override the values configured programatically with system properties but I'd need to confirm.

TL;DR I think what you're asking about is already covered by the exception type that's thrown (`SocketTimeoutException` -> `manta.tcp_socket_timeout`, `NoHttpResponseException` -> `manta.timeout`) but investigation and the request for relevant documentation has been added to #422

tjcelaya:
Merging this PR since the follow-up work has been documented in #422

PR #426, opened 2018-07-23
Download resuming/resumption/continuation

### LINKED ISSUE #411: Investigate making retries smarter by only requesting the remaining data
We'd like to explore the possibility of improving the way retries are handled. A wild idea was proposed that ~when saving to a `File`~(`File` download is implemented with `InputStream` so this has to be handled at the `InputStream` level) we should know how many bytes were successfully received and modify the retry request to specify the remaining `Range`.

For example, if a network failure occurs halfway through a download where `Content-Length: 1000`, we could add a `Range: 501-` header to indicate that we only need the second half of the file. Off the top of my head, this should be possible by manipulating the `HttpContext` provided to `MantaHttpRequestRetryHandler` to coordinate with the `MantaHttpRequestExecutor` to update the request, but how we get the value from `CountingInputStream` and hook that up with the output file is still in the air.

tjcelaya:
Resolves #411 by implementing "resumable downloads" at the `InputStream` level.

[Unit test run](https://us-east.manta.joyent.com//tomas.celaya/public/java-manta/411/Test%20Results%20-%20java-manta-client.html)
[Unit test coverage](https://us-east.manta.joyent.com//tomas.celaya/public/java-manta/411/latest-cov/index.html)

# Goal

Save bandwidth by allowing reads from files in manta to (automatically) request the remaining bytes in case of failure. Ideally implemented in a way that doesn't force users to change their calling code, or force a specific type of download (e.g. doesn't need to be a call to `getToTempFile`) since `getAsInputStream` is a commonly used method and some of our own download methods rely on it.

# Challenges

Code around the HTTP request invocations may wrap streams in helper streams (e.g. `CipherInputStream`) so coordinating method calls and replacing objects on the fly can be complicated if we intend to transparently continue reading from the source object and have already handed off the stream. Additionally, there are benefits to tracking the _total_ number of bytes read across all streams which logically represent the same object. There are also edge cases around connection pool leads, malformed or multi-valued request headers, unbounded range requests and "what happens if the failure occurs during a read that would've returned EOF".

## Handling "transparent resume"  of `InputStream`

The `ContinuingInputStream` and `AutoContinuingInputStream` classes were created specifically to handle the above issues and are described below.

## Avoiding connection pool leaks

In order to simplify resource management the class responsible for retry logic and closing errored streams (`AutoContinuingInputStream`) is separate from the class which handles tracking general progress and pluggable source streams (`ContinuingInputStream`).

## Handling "complex" range requests

This enhancement will only occur if the request being executed has "simple" values for the `If-Match` and `Range` headers. This means that either the headers are missing or present and contain a single value. If this condition is not satisfied the download will proceed normally and not be automatically resumed.

## Handling unbounded range requests

Unbounded range requests are only relevant as part of the initial request and validated against the initial response, though the initial response `Content-Range` is taken as the "target" download range.

## Handling error on EOF read

In the rare case that an `IOException` occurred while reading "past the end of the file" (i.e. the standard Java idiom of reading until EOF is returned) then we return a `ClosedInputStream` which will respond with EOF. We still attempt to close the stream which through the exception as we always would.

# Concept

The center of this PR is the `InputStreamContinuator` interface, which has the following method:

```
InputStream buildContinuation(IOException ex, long bytesRead) throws IOException;
```

i.e. when given an `IOException` that occurred while reading from another `InputStream` (be it the original or another continuation) and the total number of bytes successfully read from all related streams either:
 - rethrow the exception because it is considered *fatal*
 - build an `InputStream` which will provide the remaining bytes

# Implementation

The primary usage of this interface is through `ApacheHttpGetResponseEntityContentContinuator` whose name laboriously describes that it is an `InputStreamContinuator` for the content of an `HttpEntity` of an `HttpResponse` that was returned as the result of an `HttpGet` (and since nearly all the classes involved are from Apache HttpClient we mention Apache as well). This class requires the following inputs:
 - a ~`ResumableDownloadMarker`~ `HttpDownloadContinuationMarker` constructed from the initial request/response pair
 - an `HttpClient` or a `MantaApacheHttpClientContext` from which one may be extracted
 - the initial request which will be cloned and used for executing requests for the remainder of the object (since we update the `range` header and users may reuse their requests)

A secondary implementation of `InputStreamContinuator` was created in the form of `FileInputStreamContinuator` in order to test compatibility of the concept with `MantaEncryptedObjectInputStream`. This class adapts the metaphor for a `File` object and merely opens a new `FileInputStream` to the source file and uses `skip` to emulate starting a read from the requested offset.

### `ContinuingInputStream`/`AutoContinuingInputStream`

The `ContinuingInputStream` has been created to deal with the need for automatically swapping in a new stream (i.e. with no change to user or unrelated library code). This new stream behaves like a combination of `CountingInputStream` – Count not Continue – which counts bytes read through it and `MultipartOutputStream` which allows for swapping in a new source of data. To make retries transparent/automatic the `AutoContinuingInputStream` class which extends  `ContinuingInputStream` and wraps calls to the superclass with retry logic, automatically swapping in the new source if a continuation can be built or rethrowing the encountered exception if it was fatal.

### ~`ResumableDownloadMarker`~ `HttpDownloadContinuationMarker`

Named early on (thus the "resume" instead of "continue" terminology) and represents the `ETag`/`If-Match` and `Range`/`Content-Range` header pairs. Construction of a marker validates that the initial exchange satisfies the HTTP protocol with respect to those headers and the status code (i.e. getting a 200 response when the request was a range request is also invalid). Overall, this class performs the following functions:
 - remembers the initial `ETag` response
 - remembers the "target" range being downloaded (which may have been inferred from a `Content-Length` if the first request was not a range request)
 - verifies that subsequent responses for continuations continue to return the same ETag
 - verifies that a response content-range satisfies the current request range

### HTTP Ranges

The concept of HTTP Ranges has been formalized into the `HttpRange` class hierarchy, which has types for `UnboundedRequest`, `BoundedRequest`, and `Response` ranges. These objects can be compared to each other using the abstract `HttpRange#match` method (since equality is orthogonal) which compares all aspects that the two ranges have in common, i.e. `a.match(b)` checks all fields that `a` has are equal if present in `b`.

### `MantaEncryptedObjectInputStreamTest`

The many helper methods and helper classes originally in `MantaEncryptedObjectInputStreamTest` were extracted into `AbstractMantaEncryptedObjectInputStreamTest` so that MantaEncryptedObjectInputStreamTest only contains test methods. The helper methods were refactored slightly to allow for failure injection and verification (using mockito spies) that the continuator was used in the `MantaEncryptedObjectInputStreamAutoFailureRecoveryTest` tests.

### `HttpContextRetryCancellation`/`RetryConfigAware`

The informal "check if there's a flag in the `HttpContext` and never retry if its present" behavior of `MantaHttpRequestRetryHandler` and `MantaServiceUnavailableRetryStrategy` has been extracted into an interface named `HttpContextRetryCancellation` which provides a `default` method for this behavior. Additionally, `RetryConfigAware` was created and added to the classes which build or hold the `HttpClient`: `MantaConnectionContext` and `MantaConnectionFactory`. This allows us to completely cancel the automatic download continuation behavior in case the user passed in their own `HttpClientBuilder` since we can't reasonably check if it's safe to perform retries ourselves.

### Configuration

All the configuration classes now have a field for download continuations and the documentation has been updated to add `manta.download_continuations`/`MANTA_DOWNLOAD_CONTINUATIONS`. This value is an integer so that `-1` can represent "continue indefinitely", `0` explicitly disables continuation, and a positive integer bounds the number of allowed continuations.


java-manta-client-unshaded/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
```diff
@@ -121,6 +121,11 @@
      */
     private volatile Integer skipDirectoryDepth;
 
+    /**
+     * Whether or not we can attempt to resume a download automatically.
+     */
+    private volatile Integer downloadContinuations;
```
dekobon: Using volatile strikes me as a little odd for a non-primitive type. In particular an integer. If you were looking for a thread-safe nullable integer type, normally (for good or bad) - I tend to reach for `AtomicInteger` or a volatile primitive type. I'm not saying this is wrong, but I would double check the semantics and the need for a non-primitive type.

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
```diff
@@ -882,6 +907,7 @@ public boolean equals(final Object other) {
                 && Objects.equals(verifyUploads, that.verifyUploads)
                 && Objects.equals(uploadBufferSize, that.uploadBufferSize)
                 && Objects.equals(skipDirectoryDepth, that.skipDirectoryDepth)
+                && Objects.equals(downloadContinuations, that.downloadContinuations)
```
dekobon: It seems a bit odd that you added `downloadContinuations` to the equals() method, but not to the hashCode() method.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaHttpHeaders.java
```diff
@@ -1065,36 +1065,7 @@ public MantaHttpHeaders setByteRange(final Long start, final Long end) {
...
-            startPos = Long.parseLong(byteRange.split("-")[0]);
-        } else if (rangeParts.length == 2) {
-            startPos = Long.parseUnsignedLong(rangeParts[0]);
-            endPos = Long.parseUnsignedLong(rangeParts[1]);
-        } else {
-            throw new IllegalArgumentException("range must exist with - separator");
-        }
-
-        return new Long[] {startPos, endPos};
+        return MantaUtils.parseSingleRange(getRange());
```
dekobon: This is now calling a deprecated method. Is that the intended behavior?

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/StandardHttpHelper.java
```diff
@@ -366,28 +422,75 @@ public MantaObjectInputStream httpRequestAsInputStream(final HttpUriRequest requ
...
+        }
+
+        final HttpGet get = (HttpGet) request;
+        final ResumableDownloadMarker marker;
+        try {
+            marker = validateInitialExchange(
+                    extractDownloadRequestFingerprint(get),
+                    response.getStatusLine().getStatusCode(),
+                    extractDownloadResponseFingerprint(response, true));
+        } catch (final ProtocolException pe) {
```
dekobon: A note here about why the exception is being swallowed and null returned would be useful to the reader.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
```diff
@@ -654,13 +654,16 @@ private void readRemainingBytes() throws IOException {
...
+     * @param cipherDetails the cipher in use
      */
-    private int calculateBufferSize() {
-        long cipherTextContentLength = ObjectUtils.firstNonNull(getContentLength(), -1L);
+    static int calculateBufferSize(final Long contentLength, final SupportedCipherDetails cipherDetails) {
+        long cipherTextContentLength = ObjectUtils.firstNonNull(contentLength, -1L);
 
         if (cipherTextContentLength >= 0) {
-            cipherTextContentLength -= this.cipherDetails.getAuthenticationTagOrHmacLengthInBytes();
+            cipherTextContentLength -= cipherDetails.getAuthenticationTagOrHmacLengthInBytes();
```
dekobon: It is a little odd that you removed `this` here, but added `this` in your other changes.

dekobon:
I've added code review for style and consistency. I need to do a reread for design and I need to run the changes and test against a few failure modes.

Also, I'm curious - what happens if I set the continuations setting to a negative value?

tjcelaya:
https://github.com/joyent/java-manta/pull/426/files#diff-0f23af2f696313fe32c17a0c7281cb57R168

`-1` enables unlimited continuations, `0` disables, positives enabled bounded continuations (a failsafe in case of some unexpected loop)

> I've added code review for style and consistency

Don't see any comments on the PR?

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
tjcelaya: This method was made static and package-private so that it can be used from `MantaEncryptedObjectInputStreamAutoFailureRecoveryTest` to inject failures relative to the buffer size: https://github.com/joyent/java-manta/pull/426/files#diff-58132a0f33da576d1b7e967879396e57R230

dekobon: :+1: I didn't catch that detail from the diff. Thanks!

dekobon: Oh, if it is static, then we should use the class name to access it. That's what was tripping up my eyes.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/MantaHttpHeaders.java
tjcelaya: The method was extracted into `MantaUtils` early on in the project but abandoned in favor of the `HttpRange` implementation. I'm putting the method body back and updating the JavaDoc to explain the "soft deprecation."

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
tjcelaya: Silly mistake, sorry.

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
tjcelaya: I was following the existing pattern based on our decision in [this comment](https://github.com/joyent/java-manta/pull/357#issuecomment-340869194) when we made the ConfigContext reloadable. I opted for an `Integer` since the default was was originally `null` for "disable continuations" but since we have an explicit off value (`0`) now we can make it a primitive `int`.

pom.xml
```diff
@@ -507,7 +507,18 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-maven-plugin</artifactId>
```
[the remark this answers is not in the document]
tjcelaya: Fixed, code convention/style != nitpick

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/crypto/MantaEncryptedObjectInputStream.java
tjcelaya: Added class name to clarify static access.

java-manta-client-unshaded/src/main/java/com/joyent/manta/http/StandardHttpHelper.java
tjcelaya: Added a debug log which includes the reason we couldn't build a marker

tjcelaya:
To make the behavior very obvious, I've included below a range download starting halfway into the file and manually terminated the request twice during the download. Trimmed down the headers to only show relevant ones. The test case slices the plaintext object and compares that the bytes read from the stream match the same byte range of the original file. Additionally JMX was being monitored to see what kind of exception was triggered (thus the `get-continuations-recovered-exception-ConnectionClosedException` metric reaching `2`).

(Initial request)

  | GET /tomas.celaya/stor/java-manta-integration-tests/1fb63d00-0fbb-4300-8f0a-d4449128bc07/ApacheHttpGetResponseEntityContentContinuatorIT//25bab7f5-194f-45f6-af0d-06a2fca9d136 HTTP/1.1
-- | --
x-request-id | 9d8fa9f3-942c-11e8-a0cd-51bd36c69e70
Date | Mon, 30 Jul 2018 19:13:17 GMT
range | bytes=446453-
User-Agent | Java-Manta-SDK/3.2.3-SNAPSHOT (Java/1.8.0_131/Oracle Corporation)

(Initial response)

  | HTTP/1.1 206 Partial Content
-- | --
Etag | 00a39773-7ab8-40a9-9b4f-9ccb23f0946d
Last-Modified | Mon, 30 Jul 2018 19:12:47 GMT
Content-Type | application/octet-stream
Content-Range | bytes 446453-892906/892907
Content-Length | 446454
Durability-Level | 2
Date | Mon, 30 Jul 2018 19:13:18 GMT
x-request-id | 9d8fa9f3-942c-11e8-a0cd-51bd36c69e70
Connection | Keep-alive

(continuator-initiated request)


  | GET /tomas.celaya/stor/java-manta-integration-tests/1fb63d00-0fbb-4300-8f0a-d4449128bc07/ApacheHttpGetResponseEntityContentContinuatorIT//25bab7f5-194f-45f6-af0d-06a2fca9d136 HTTP/1.1
-- | --
x-request-id | 9d8fa9f3-942c-11e8-a0cd-51bd36c69e70
Date | Mon, 30 Jul 2018 19:13:17 GMT
Range | bytes=639709-892906
If-Match | 00a39773-7ab8-40a9-9b4f-9ccb23f0946d
User-Agent | Java-Manta-SDK/3.2.3-SNAPSHOT (Java/1.8.0_131/Oracle Corporation)
Accept-Encoding | gzip,deflate

(continuator response)

  | HTTP/1.1 206 Partial Content
-- | --
Etag | 00a39773-7ab8-40a9-9b4f-9ccb23f0946d
Last-Modified | Mon, 30 Jul 2018 19:12:47 GMT
Content-Type | application/octet-stream
Content-Range | bytes 639709-892906/892907
Content-Length | 253198
Durability-Level | 2
Date | Mon, 30 Jul 2018 19:13:25 GMT
Server | Manta
x-request-id | 9d8fa9f3-942c-11e8-a0cd-51bd36c69e70
Connection | Keep-alive

(second continuator-initiated request)

  | GET /tomas.celaya/stor/java-manta-integration-tests/1fb63d00-0fbb-4300-8f0a-d4449128bc07/ApacheHttpGetResponseEntityContentContinuatorIT//25bab7f5-194f-45f6-af0d-06a2fca9d136 HTTP/1.1
-- | --
x-request-id | 9d8fa9f3-942c-11e8-a0cd-51bd36c69e70
Date | Mon, 30 Jul 2018 19:13:17 GMT
Range | bytes=700445-892906
If-Match | 00a39773-7ab8-40a9-9b4f-9ccb23f0946d
User-Agent | Java-Manta-SDK/3.2.3-SNAPSHOT (Java/1.8.0_131/Oracle Corporation)
Accept-Encoding | gzip,deflate

(second continuator-initiated response)

  | HTTP/1.1 206 Partial Content
-- | --
Etag | 00a39773-7ab8-40a9-9b4f-9ccb23f0946d
Last-Modified | Mon, 30 Jul 2018 19:12:47 GMT
Content-Type | application/octet-stream
Content-Range | bytes 700445-892906/892907
Content-Length | 192462
Durability-Level | 2
Date | Mon, 30 Jul 2018 19:13:29 GMT
Server | Manta
x-request-id | 9d8fa9f3-942c-11e8-a0cd-51bd36c69e70
Connection | Keep-alive

























Screenshot of the continuator in action (manually terminating requests in CWP on the left, JMX showing recovery count on the right):

screen shot 2018-07-30 at 12 14 10 pm

tjcelaya:
Application logs while running `ApacheHttpGetResponseEntityContentContinuatorIT#regularObjectDownloadEncrypted` can be seen at [in this gist](https://gist.github.com/tjcelaya/df6a618a4d89fa96abe65e0cc3354385#file-encrypted-object-download-log).

tjcelaya:
Application logs while running `ApacheHttpGetResponseEntityContentContinuatorIT#rangeObjectDownloadEncrypted` can be seen in [this gist](https://gist.github.com/tjcelaya/df6a618a4d89fa96abe65e0cc3354385#file-encrypted-range-download-log).

tjcelaya:
The JavaDoc for the integration test class is reformatted and pasted below:

This class encapsulates the situations meant to be remedied by {@link com.joyent.manta.http.ApacheHttpGetResponseEntityContentContinuator} for manual verification.

Until the issue with WireMock is resolved or Charles Web Proxy allows dynamic request termination through its control panel integration testing of `com.joyent.manta.http.ApacheHttpGetResponseEntityContentContinuator` will require manual user intervention. There is more griping about this (and a Github issue link) in the final paragraph of this JavaDoc.

Users are expected to have access to [Charles Web Proxy](https://www.charlesproxy.com/) or some other proxy which allows them to to cause request failures in a controlled fashion. This test specifically requires turning on throttling so that the download is artificially slowed (giving the tester time to react) and then manually cancelling GET requests in order to trigger the continuator. Warning log messages are written out in a very obvious fashion indicating when the test is ready to progress and when user intervention is no longer required. If the test operator does not terminate any requests the test will fail when it notices no metrics were recorded.

Since we want to make it as easy as possible to run this test without a code change this test throws a `org.testng.SkipException` if the proxy settings are missing. We expect to find the following system properties to not be blank:

*   http.proxyHost
*   http.proxyPort
*   https.proxyHost
*   https.proxyPort

Also note that since several other test cases cover the interactions between `com.joyent.manta.util.InputStreamContinuator` and `com.joyent.manta.client.crypto.MantaEncryptedObjectInputStream` we only test with a single key strength for each cipher.

*Recommended throttle preset for Charles Web Proxy*: 256K ISDN/DSL
(this causes each download to take ~30 seconds)

Recommended testing procedure:

*   Start Charles Web Proxy locally
*   Run the test class or an individual test method, making sure to pass the aforementioned system properties
*   As a sanity check, this class uploads and downloads one file for each cipher tested before actually starting any test methods. The use of a "null" cipher means no encryption.
*   Watch the logs for ` >>> Pausing for 15s....` which will occur before each test method is run. This gives the tester time to clear the Charles Web Proxy request log and enable throttling.
*   Once the actual request is in flight, (logs show ` --- Starting [plain|range] [encrypted|unencrypted] download`) the tester should manually terminate the next GET request that shows up in Charles Web Proxy. This can be done as many times as desired.
*   Assuming at least a single request was forcefully terminated, the test case should pass. If the tester forgets to terminate a request, an AssertionError will be thrown when the test sees that no continuation metrics were recorded.

We would ideally automate this with WireMock but there is currently [no way to configure](https://github.com/tomakehurst/wiremock/issues/966) it to simulate failures exclusively in the response body. Similarly, while Charles Web Proxy allows users to enable a control interface over HTTP it provides no way to terminate in-flight requests.

PR #428, opened 2018-07-27
Allow users to pass in MantaHttpHeaders with MantaClient#delete

tjcelaya:
API changes:
 - package-private constructor for `MantaClient` to allow passing an `HttpHelper` directly
 - method `MantaClient#delete(String, MantaHttpHeaders)` added
 - method `HttpHelper#httpDelete(String, MantaHttpHeaders)` added
 - public constructor for `MantaClientHttpResponseException(HttpRequest, HttpResponse, String, Integer...)` added (in case the response is <400 and not one of the valid DELETE response codes: 200/202/204)

Unit tests were added for:
- `MantaClient#delete` → `HttpHelper#httpDelete`, verifying the `If-Match` header passed in is seen by the helper
- `HttpHelper` → `HttpClient`, verifying the request seen by the `HttpClient` contains the expected header value

Integration test added:
 - create an empty file
 - grab its ETag
 - attempt to delete with an invalid ETag (string reverse of ETag, the behavior of `If-Match: ""` seems to be undefined)
   - make sure we get the expected exception (412 Precondition Failed response code)
   - make sure the object still exists
- delete with correct ETag in `If-Match` header
   - make sure the object no longer exists

Bystanders:
 - minor `@Test` annotation cleanup
 - extracted common path generation code in one of the unit tests into the `generatePath` helper. might belong in a more general class but left in its own test class for now

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaClientHttpResponseException.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaClientHttpResponseException.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 
 /**
  * Exception class representing a failure in the contract of Manta's behavior.
@@ -126,7 +127,22 @@ public class MantaClientHttpResponseException extends MantaIOException {
     public MantaClientHttpResponseException(final HttpRequest request,
                                             final HttpResponse response,
                                             final String path) {
-        super(String.format("HTTP request failed to: %s", path));
+        this(request, response, path, (Integer[]) null);
+    }
+
+    /**
+     * Builds a client exception object that is annotated with all of the
+     * relevant request and response debug information.
+     *
+     * @param request HTTP request object
+     * @param response HTTP response object
+     * @param path The fully qualified path of the object. i.e. /user/stor/foo/bar/baz
+     */
+    public MantaClientHttpResponseException(final HttpRequest request,
+                                            final HttpResponse response,
+                                            final String path,
+                                            final Integer... expectedResponseCodes) {
+        super(buildExceptionMessageFromHttpExchange(request, response, path, expectedResponseCodes));
         final HttpEntity entity = response.getEntity();
         final ContentType jsonContentType = ContentType.APPLICATION_JSON;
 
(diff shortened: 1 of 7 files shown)


java-manta-it/src/test/java/com/joyent/manta/http/MantaHttpHeadersIT.java
```diff
@@ -270,4 +272,36 @@ public void canSetDurability() throws IOException {
         Assert.assertEquals(actualHeaders.getDurabilityLevel().intValue(), durability,
                 "Durability not set to value on put");
     }
+
+    public void canFailToDeleteDirectoryOnBadIfMatch() throws IOException {
```
melloc: Who uses this? As far as I can tell, there are no consumers.

tjcelaya: I'm not sure if you're asking about the `canFailToDeleteDirectoryOnBadIfMatch` method or the usage of `if-match`? JUnit/TestNG executes all public methods in a test class. The way it's designated as a test class is because it's referenced by [this line](https://github.com/joyent/java-manta/blob/master/java-manta-it/src/test/resources/testng-it.xml#L26) in the testng.xml configuration.

(`MantaClientTest#deleteWithHeadersPassesAlongHeaders` and the methods added to `StandardHttpHelperTest` are picked up the same way but with a different testng.xml)

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaClientHttpResponseException.java
```diff
@@ -183,6 +200,28 @@ public MantaClientHttpResponseException(final HttpRequest request,
...
+                                                                final HttpResponse response,
+                                                                final String path,
+                                                                final int... expectedResponseCodes) {
+        if (expectedResponseCodes != null) {
+            return String.format("HTTP request returned unexpected response code: expected one of %s, got [%d] ",
+                                 Arrays.toString(expectedResponseCodes),
+                                 response.getStatusLine().getStatusCode());
+        }
+
+        return String.format("HTTP request failed to: %s", path);
```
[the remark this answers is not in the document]
tjcelaya: I was extracting a method and only altering the behavior based on the new inputs, the generic case may or may not be relevant to the status code so I didn't modify the standard message (i.e. any callers not passing in a list of allowed status codes will still produce the same exception message as before).

java-manta-it/src/test/java/com/joyent/manta/http/MantaHttpHeadersIT.java
melloc: Ah, I see. I was wondering how the method gets called. I didn't realize that JUnit would take care of discovering and executing it, so this makes more sense to me now.

tjcelaya:
@melloc @dekobon 

approved for merge?

melloc:
@tjcelaya Looks good to me! (I thought that there was an actual approve button, but now I can't find it.)

PR #429, opened 2018-08-02
Improve AutoContinuingInputStream failure case

tjcelaya:
Currently if an unrecoverable exception happens while using try-with-resources in combination with auto-resume the following will occur:

- `AutoContinuingInputStream` will catch it and invoke `attemptRecovery` to ask its continuator for a continuation
- the continuator will rethrow the exception because it is considered fatal
- immediately `ContinuingInputStream` will call `this.discardWrapped()` which closes the wrapped stream and sets `this.wrapped` to `null`
- the catch in `attemptRecovery` will attempt to add the newly caught exception as a suppressed exception of the original (thinking it was related to recovery)
- try-with-resources sees the original exception and calls close on the outer stream which triggers an NPE when `AutoContinuingInputStream` calls `this.getWrapped().close()` (because the superclass already discarded the wrapped stream so there is nothing to close) so that NPE is attached to the original exception as a suppressed exception

This is all much more complicated than it needs to be and is the result of over-eager behavior on the part of `ContinuingInputStream` and a missing null check in `AutoContinuingInputStream#close`. The resulting exception looks like this:

```
java.lang.IllegalArgumentException: Self-suppression not permitted
	at java.lang.Throwable.addSuppressed(Throwable.java:1043)
	at com.joyent.manta.util.AutoContinuingInputStream.attemptRecovery(AutoContinuingInputStream.java:59)
	at com.joyent.manta.util.AutoContinuingInputStream.read(AutoContinuingInputStream.java:81)
	at com.joyent.manta.client.MantaObjectInputStream.read(MantaObjectInputStream.java:175)
	...
	at org.apache.commons.io.IOUtils.toByteArray(IOUtils.java:721)
	at com.joyent.manta.client.ApacheHttpGetResponseEntityContentContinuatorIT.regularObjectDownloadUnencrypted(ApacheHttpGetResponseEntityContentContinuatorIT.java:244)
	Suppressed: java.lang.NullPointerException
		at com.joyent.manta.util.AutoContinuingInputStream.close(AutoContinuingInputStream.java:138)
		at com.joyent.manta.org.apache.commons.io.IOUtils.closeQuietly(IOUtils.java:339)
		at com.joyent.manta.org.apache.commons.io.IOUtils.closeQuietly(IOUtils.java:270)
		at com.joyent.manta.client.MantaObjectInputStream.close(MantaObjectInputStream.java:190)
		at com.joyent.manta.client.ApacheHttpGetResponseEntityContentContinuatorIT.$closeResource(ApacheHttpGetResponseEntityContentContinuatorIT.java:231)
		at com.joyent.manta.client.ApacheHttpGetResponseEntityContentContinuatorIT.regularObjectDownloadUnencrypted(ApacheHttpGetResponseEntityContentContinuatorIT.java:245)
		... 30 more
Caused by: javax.net.ssl.SSLException: SSL peer shut down incorrectly
	at sun.security.ssl.InputRecord.readV3Record(InputRecord.java:596)
	...
	at com.joyent.manta.org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:148)
	at com.joyent.manta.util.ContinuingInputStream.read(ContinuingInputStream.java:146)
	at com.joyent.manta.util.AutoContinuingInputStream.read(AutoContinuingInputStream.java:79)
	... 37 more
```

Instead, we'd rather just throw `javax.net.ssl.SSLException: SSL peer shut down incorrectly` and call it a day. This PR fixes that situation and adds a test for verifying that an exception rethrown by `InputStreamContinuator#buildContinuation` is rethrown as-is.

This PR also makes `ContinuingInputStream` easier to use by removing the automatic calls to             `discardWrapped()` on any exception so some redundant `try` statements have also been removed.

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
@@ -238,7 +238,7 @@ public class ApacheHttpGetResponseEntityContentContinuator implements InputStrea
                           + "to recover at byte offset {} "
                           + "from exception {}",
                   this.request.getMethod(),
-                  this.request.getRequestLine(),
+                  this.request.getRequestLine().getUri(),
                   bytesRead,
                   ex.getMessage());
 
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
@@ -232,10 +232,15 @@ public class ApacheHttpGetResponseEntityContentContinuator implements InputStrea
                     ex);
         }
 
-        LOG.debug("Attempting to build a continuation for request {} to recover at byte offset {} from exception {}",
+        LOG.debug("Attempting to build a continuation for "
+                          + "[{}] request "
+                          + "to path [{}] "
+                          + "to recover at byte offset {} "
+                          + "from exception {}",
+                  this.request.getMethod(),
                   this.request.getRequestLine(),
                   bytesRead,
-                  ex);
+                  ex.getMessage());
 
         // if an IOException occurs while reading EOF the user may ask us for a continuation
         // starting after the last valid byte.
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
@@ -218,6 +218,8 @@ public class ApacheHttpGetResponseEntityContentContinuator implements InputStrea
      */
     @Override
     public InputStream buildContinuation(final IOException ex, final long bytesRead) throws IOException {
+        requireNonNull(ex);
+
         if (!isRecoverable(ex)) {
             throw ex;
         }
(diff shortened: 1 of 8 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/http/ApacheHttpGetResponseEntityContentContinuator.java
```diff
@@ -232,10 +232,15 @@ public InputStream buildContinuation(final IOException ex, final long bytesRead)
...
+        LOG.debug("Attempting to build a continuation for "
+                          + "[{}] request "
+                          + "to path [{}] "
+                          + "to recover at byte offset {} "
+                          + "from exception {}",
+                  this.request.getMethod(),
+                  this.request.getRequestLine().getUri(),
                   bytesRead,
-                  ex);
+                  ex.getMessage());
```
dekobon: Is there any possibility of `ex` being null?

java-manta-client-unshaded/src/main/java/com/joyent/manta/util/AutoContinuingInputStream.java
```diff
@@ -54,9 +58,16 @@ public AutoContinuingInputStream(final InputStream wrapped,
      */
     private void attemptRecovery(final IOException originalIOException) throws IOException {
         try {
-            this.continueWith(this.continuator.buildContinuation(originalIOException, this.getBytesRead()));
-        } catch (final IOException ce) {
-            originalIOException.addSuppressed(ce);
+            super.continueWith(this.continuator.buildContinuation(originalIOException, this.getBytesRead()));
+        } catch (final IOException ioe) {
```
dekobon: This might be paranoid, but you may want to catch `UncheckedIOException` as well.

Issue #425, opened 2018-07-19
Provide a way to automatically prune empty parent directories


tjcelaya:
In certain deployments it makes sense to clean up as many parent directories as possible after deleting an object, a sort of reverse `deleteRecursive`. Assuming the following directory structure:

```
/$USER/stor/dir1
/$USER/stor/dir1/dir2
/$USER/stor/dir1/dir2/dir3A
/$USER/stor/dir1/dir2/dir3A/object
/$USER/stor/dir1/dir2/dir3B
```

The basic algorithm would be described as follows.

```
user calls:
mantaClient.delete("/$USER/stor/dir1/dir2/dir3A/object");

required http calls:
DELETE /$USER/stor/dir1/dir2/dir3A/object  204 No Content

pruning http calls:
DELETE /$USER/stor/dir1/dir2/dir3A         204 No Content
DELETE /$USER/stor/dir1/dir2               400 Bad Request (DirectoryNotEmpty)
```

This could be implemented either as:
 - new method: `MantaClient#delete(String path, boolean prune)` for unbounded pruning
 - new method: `MantaClient#delete(String path, int prune)` for bounded pruning
 - new setting: `manta.prune_empty_parents_after_delete=[ Integer ]`  which indicates regular calls to `delete` (and/or `deleteRecursive`) should also trigger pruning (bounded by the integer). 
 - new setting: `manta.prune_empty_parents_after_delete=[ Boolean ]`  which indicates regular calls to `delete` (and/or `deleteRecursive`)  should also trigger pruning (unbounded). 

The difference between boolean and integer parameters refer to unbounded pruning (i.e. deleting all non-empty parents) while an integer option would instead limit the number of cleanup operations (and perhaps use `-1` as an unbounded option).

PR #431, opened 2018-08-08
Fix/425

douglasAtJoyent:
This is for Issue #425, I have added all of the code, this adds a parameter MANTA_PRUNE_EMPTY_PARENT_DEPTH, and if this is set will attempt to delete up to this depth of empty parent directories. If given a negative number it will try to delete all empty parent directories.

I am still working on the documentation, I will have it checked in shortly, I just wanted to get the code review started sooner than later.


java-manta-client-unshaded/src/main/java/com/joyent/manta/client/PruneEmpytParentDirectoryStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.joyent.manta.exception.MantaClientHttpResponseException;
+import com.joyent.manta.exception.MantaErrorCode;
+import com.joyent.manta.http.MantaHttpHeaders;
+
+/**
+ * Utility class for recursive directory creation strategies.
+ *
+ * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
+ * @since 3.1.7
```
tjcelaya: `@author` should be you `@since` should be 3.2.4

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -358,14 +358,29 @@ public void delete(final String rawPath) throws IOException {
...
+        }
+    }
+
+    /**
+     * Deletes several directories in manta.
+     * @param paths - A list of paths to be deleted.
+     * @param requestHeaders - headers can be null.
+     * @throws IOException If a HTTP status code other than {@code 200 | 202 | 204} is encountered
+     */
+    public void delete(final List<String> paths, final MantaHttpHeaders requestHeaders) throws IOException {
```
tjcelaya: I wouldn't recommend increasing the API surface for a convenience method like this. The potential failure modes for this method mean that users can't be sure which delete in the list failed.

I'd venture to guess you're trying to avoid the infinite recursion of `MantaClient#delete(String, MantaHttpHeaders)` and `PruneEmpytParentDirectoryStrategy#pruneParentDirectories(String)` calling each other indefinitely. If that's the case I'd recommend adding a package-private `MantaClient#delete(String, MantaHttpHeaders, Integer)` method which can perform a regular delete when the `Integer` argument is null and delegates to the strategy class otherwise. Then the strategy class can explicitly pass `null` to get the plain single-delete functionality. In that case the body of `MantaClient#delete(String, MantaHttpHeaders)` would just be `this.delete(path, headers, config.getPruneEmptyParentDepth())`

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/RecursiveDirectoryCreationStrategy.java
```diff
@@ -20,8 +20,8 @@
 /**
  * Utility class for recursive directory creation strategies.
  *
- * @author <a href="https://github.com/tjcelaya">Tomas Celayac</a>
- * @since 3.1.7
+ * @author <a href="hhttps://github.com/douglasAtJoyent">Douglas Anderson</a>
+ * @since 3.2.4
```
tjcelaya: I think you mixed up RecursiveDirectoryCreationStrategy.java and PruneEmpytParentDirectoryStrategy.java when making this change

douglasAtJoyent: Sorry, will revert and change the correct file.

USAGE.md
```diff
@@ -168,6 +169,10 @@ Note: Dynamic Updates marked with an asterisk (*) are enabled by the `AuthAwareC
 * `manta.skip_directory_depth` (**MANTA_SKIP_DIRECTORY_DEPTH**)
     Integer indicating the number of **non-system** directory levels to attempt to skip for recursive `putDirectory`
     operation (i.e. `/$MANTA_USER` and `/$MANTA_USER/stor` would not be counted). A detailed explanation and example are provided [later in this document](/USAGE.md#skipping-directories)
+* `manta.prune_empty_parent_depth` (**MANTA_PRUNE_EMPTY_PARENT_DEPTH**)
+    Integer indicating the maximum number of empty parent directories to prune. If the client is given the value of -1 then the client will try and delete
+    all empty parent directories.
+    USAGE.md#Pruning empty parent directories)
```
tjcelaya: `USAGE.md#Pruning empty parent directories)`

looks like a typo, I assume this was supposed to be something like

`see [the section on parent directory pruning later in this document](/USAGE.md#Pruning empty parent directories)`

USAGE.md
```diff
@@ -537,6 +542,29 @@ In order to ease migration from other object stores which do not treat directori
...
+   If you have prune_empty_parent_depth set to 1 then delete test.txt, the client should delete Dir3 as well.
+ 
+#### Scenario 2 : Prune Empty Parent Depth -1
+- Given the directory structure : 
+   /Dir1
+   /Dir1/DirA
+   /Dir1/Dir2
+   /Dir1/Dir2/Dir3
+   /Dir1/Dir2/Dir3/test.txt
+   In this case this will if test.txt this will delete test.txt, dir3, and Dir2. It will stop when it gets to the Dir1 since it will not be empty.
```
tjcelaya: > In this case this will if test.txt this will delete test.txt, dir3, and Dir2.`

=> 

> In this case deletion of `test.txt` will also delete `Dir3`, and `Dir2`.

douglasAtJoyent: Nope, test.txt is the target, the parent is Dir3, and with the depth of 1 it will only delete Dir3 and not Dir2.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -358,15 +358,27 @@ public void delete(final String rawPath) throws IOException {
...
-        Validate.notBlank(rawPath, "rawPath must not be blank");
+    public void delete(final String rawPath, final MantaHttpHeaders requestHeaders) throws IOException {
+        delete(rawPath, requestHeaders, config.getPruneEmptyParentDepth());
+    }
 
+    /**
+     * Deletes several directories in manta.
+     * @param paths - A list of paths to be deleted.
+     * @param requestHeaders - headers can be null.
+     * @throws IOException If a HTTP status code other than {@code 200 | 202 | 204} is encountered
```
tjcelaya: Missing `@param pruneDepth`.

It seems checkstyle is ignoring JavaDocs for package-private methods, I'll see how we can fix that.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/PruneEmpytParentDirectoryStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+    static void pruneParentDirectories(final MantaClient client,
+            final MantaHttpHeaders headers,
+            final String path,
+            final int limit) throws IOException {
+        // First thing first, delete the child directory.
+        client.delete(path, headers, null);
+
+        int actualLimit = limit;
+        if (actualLimit == PRUNE_ALL_PARENTS) {
+            actualLimit = path.split(MantaClient.SEPARATOR).length - 1;
```
tjcelaya: I'd recommend saving the value of `path.split(MantaClient.SEPARATOR)` outside this check you can reuse it later in the for loop.

Also: doesn't this end up counting the `/USER` and `/USER/stor` directories as well? `MantaUtils#writeablePrefixPaths` will give you all the paths the user can actually write to (passed-in path).

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/ConfigContext.java
```diff
@@ -248,6 +254,7 @@ static String toString(final ConfigContext context) {
         sb.append(", verifyUploads=").append(context.verifyUploads());
         sb.append(", uploadBufferSize=").append(context.getUploadBufferSize());
         sb.append(", skipDirectoryDepth=").append(context.getSkipDirectoryDepth());
+        sb.append(", pruneEmptyParentDepth=").append(context.getPruneEmptyParentDepth());
         sb.append(", metricReporterMode=").append(context.getMetricReporterMode());
```
tjcelaya: I noticed the downloadContinuations setting wasn't being output here so I edited it in master. That _might_ cause a merge conflict (which I can resolve through the PR).

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/ConfigContext.java
```diff
@@ -141,6 +141,12 @@ default String getMantaHomeDirectory() {
      */
     Integer getSkipDirectoryDepth();
 
+    /**
+     * @return the maximum amount of empty parent directories to delete. -1 will delete all empty parent directories 0
+     *         will delete none of them.
+     */
+    Integer getPruneEmptyParentDepth();
```
tjcelaya: I can only add comments within changed areas but this setting should also be checked in the `validate` method. Search for "Manta skip directory depth must be 0 or greater" (should be around line 344 according to this diff)

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/SettableConfigContext.java
```diff
@@ -180,6 +180,14 @@
      */
     T setSkipDirectoryDepth(Integer depth);
 
+    /**
+     * Sets a limit to the number of parent directories deleted, after a delete.
+     *
+     * @param depth directory depth to assume exists
```
tjcelaya: Holdover from `setSkipDirectoryDepth`, should say something like `empty parents to prune upon deletion of an object`

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/PruneEmpytParentDirectoryStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+        LOG.debug("Actual Number : " + actualLimit + " of directories");
+        // I am generating all of the possible paths first, then we will delete them.
+        String parent = path;
+        try {
+            for (int i = 0; i < actualLimit; i++) {
+                parent = parent.substring(0, parent.lastIndexOf("/"));
+                client.delete(parent, headers, null);
+            }
+        } catch (MantaClientHttpResponseException responseException) {
+            if (responseException.getServerCode() != MantaErrorCode.DIRECTORY_NOT_EMPTY_ERROR) {
```
tjcelaya: I would recommend adding another check before this one:

```
            // this directory may have been deleted by another process, continue to the parent
            if (e.getServerCode().equals(MantaErrorCode.RESOURCE_NOT_FOUND_ERROR)) {
                continue;
            }
```

java-manta-client-unshaded/src/main/java/com/joyent/manta/config/BaseChainedConfigContext.java
```diff
@@ -121,6 +121,13 @@
      */
     private volatile Integer skipDirectoryDepth;
 
+    /**
+     * Clean up as many parent directories as possible after deleting an object.
+     * This is the greatest number of directories that will be attempted
+     * to be deleted.
+     */
+    private volatile Integer pruneEmptyParentDepth;
```
tjcelaya: Please also add this to the properties checked in the `equals` method.

USAGE.md
tjcelaya: Wouldn't the deletion of `Dir3` mean that Dir2 is empty (the scenario says `-1` so I'd assume it would continue up to `Dir1` and stop there because of `DirA`)

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/PruneEmpytParentDirectoryStrategy.java
douglasAtJoyent: I thought about that, but I thought it would be a better idea to let the exception raise if there is a directory in a path that was supplied, but is not longer found. I can imagine that may happen if there are multiple threads trying to delete things from the same place, but then I think it should just return and not continue.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientDirectoriesIT.java
```diff
@@ -242,4 +242,93 @@ private void directorySkipping(final boolean settingCorrectDepth) throws IOExcep
...
+
+    @Test
+    public void pruneParentDirectoryOne() throws IOException {
+        // This should stop at 1 parent being deleted
+        final String parentDir = createRandomDirectory(testPathPrefix, 1);
+        final String childDir = createRandomDirectory(parentDir, 3);
+        LOG.info("CHILD DIR  : " + childDir);
+        LOG.info("Parent DIR : " + parentDir);
+        mantaClient.delete(childDir, null, 1);
+        MantaAssert.assertResponseFailureStatusCode(404, RESOURCE_NOT_FOUND_ERROR,
```
tjcelaya: Not a necessary change but these can probably be simplified to `Assert.assertFalse(mantaClient.existsAndIsAccessible(childDir)`.

Up to you if you want to make this change.

java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaClient.java
```diff
@@ -380,15 +380,27 @@ public void delete(final String rawPath) throws IOException {
...
+     */
+    void delete(final String rawPath, final MantaHttpHeaders requestHeaders, final Integer pruneDepth)
+            throws IOException {
+        Validate.notBlank(rawPath, "rawPath must not be blank");
         String path = formatPath(rawPath);
-        LOG.debug("DELETE {}", path);
-
-        httpHelper.httpDelete(path, requestHeaders);
-    }
+        if (pruneDepth == null) {
```
tjcelaya: It's good that you can handle the 0 case in `PruneEmpytParentDirectoryStrategy` but it's easier on the next maintainer if you make this check `pruneDepth == null || pruneDepth == 0` (since the fact that it isn't null means the unboxing will be safe)

PR #434, opened 2018-10-04
Adding fix for integration test to work with different behavior of the

douglasAtJoyent:
Changing the test for pruneParentDirectoriesInvalid. Previously the client would ignore any value of -1 for the prune depth. Now it throws an exception. 

This test uses -3, so now the test should result in an exception being thrown.

Fixing issue 433.


nairashwin952013:
Received will check shortly!

PR #447, opened 2018-12-05
Connection timeout error message removed for test class TCPSocketConnectionTimeoutIT

nairashwin952013:
Java uses different underlying network libraries between Linux, MacOS, SmartOS and Windows, so the message returned in exceptions like the following are not consistent. Hence this ```assertTrue()``` statement had to be removed to make the test consistent for all systems.

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/http/TCPSocketConnectionTimeoutIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/http/TCPSocketConnectionTimeoutIT.java
@@ -62,12 +62,10 @@ public class TCPSocketConnectionTimeoutIT {
             start = Instant.now();
             client.head(testPathPrefix);
 
-        } catch(InterruptedIOException e){
+        } catch (InterruptedIOException e) {
             Assert.assertEquals(e.getClass().getSimpleName(),
-                    "ConnectTimeoutException");
-            Assert.assertTrue(e.getMessage().endsWith("connect timed out"),
-                    "ConnectTimeoutException didn't end with expected "
-                            + "connection timeout message. Actual exception:\n"
+                    "ConnectTimeoutException",
+                    "Expected exception ConnectTimeoutException was not thrown. Actual Exception:\n"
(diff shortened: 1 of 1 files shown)


java-manta-it/src/test/java/com/joyent/manta/http/TCPSocketConnectionTimeoutIT.java
```diff
@@ -65,10 +65,7 @@ public void verifyConnectionTimeoutSettingWorks() throws IOException {
         } catch(InterruptedIOException e){
             Assert.assertEquals(e.getClass().getSimpleName(),
                     "ConnectTimeoutException");
-            Assert.assertTrue(e.getMessage().endsWith("connect timed out"),
-                    "ConnectTimeoutException didn't end with expected "
-                            + "connection timeout message. Actual exception:\n"
-                            + ExceptionUtils.getStackTrace(e));
+            ExceptionUtils.getStackTrace(e);
```
dekobon: This line makes no sense. What did you expect this to do and why?

java-manta-it/src/test/java/com/joyent/manta/http/TCPSocketConnectionTimeoutIT.java
```diff
@@ -64,11 +64,7 @@ public void verifyConnectionTimeoutSettingWorks() throws IOException {
 
         } catch(InterruptedIOException e){
             Assert.assertEquals(e.getClass().getSimpleName(),
-                    "ConnectTimeoutException");
-            Assert.assertTrue(e.getMessage().endsWith("connect timed out"),
-                    "ConnectTimeoutException didn't end with expected "
-                            + "connection timeout message. Actual exception:\n"
-                            + ExceptionUtils.getStackTrace(e));
+                    "ConnectTimeoutException", ExceptionUtils.getStackTrace(e));
```
dekobon: Someone who sees this test failure isn't going to understand why the stacktrace was printed. It would be better to provide a message like: "Expected exception ConnectTimeoutException was not thrown. Actual Exception: %s"

PR #464, opened 2019-01-15
Modifying testng-it.xml file to exclude New Run-Config

nairashwin952013:
These changes were introduced to exclude the newly made ```MultiCryptoProviderDecryption``` unit test from certain tests specified in the ```testng-it.xml``` file. It will help avoid failures if we source this xml file to run a test-suite.

### DIFF
--- a/java-manta-it/src/test/resources/testng-it.xml
+++ b/java-manta-it/src/test/resources/testng-it.xml
@@ -16,6 +16,7 @@
         <groups>
             <run>
                 <exclude name="encrypted"/>
+                <exclude name="encryption-provider" />
             </run>
         </groups>
         <packages>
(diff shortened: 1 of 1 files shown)


nairashwin952013:
While sourcing the test-suite:```testng-it.xml``` file it automatically includes the new unit-test made to test multiple-providers since its included in the package ```com.joyent.manta.client.*``` mentioned in the run-configuration for the three test names: i.e ```Manta Client Integration Tests [AES128/CTR Encrypted], Manta Client Integration Tests [AES128/GCMEncrypted] & Manta Client Integration Tests [AES128/CBC Encrypted]``` so the Suite was failing since the proper invocation parameters for that particular test class aren't invoked since we need to specify the provider in order to run that test class. 

Hence, by excluding the group encryption-provider we can safely ensure that we get a fully-green test suite while running a test-configuration where you only specify the test name using ```-testname Manta Client Integration Tests [AES128/CTR Encrypted]``` .

PR #469, opened 2019-01-17
Changed Disabled mode to VerificationDisabled Mode

nairashwin952013:
Since we want to disable both the verification and creation of HMAC signatures for authenticating ciphertext. The enum value that was created (to disable ciphertext verification upon decryption where HMACs will still be created while encrypting. ) in [PR#466](https://github.com/joyent/java-manta/pull/466/files) defined ```Disabled``` has been modified to be called ```VerificationDisabled```.


nairashwin952013:
All checks have passed, Merging [PR#469](https://github.com/joyent/java-manta/pull/469)

PR #484, opened 2019-02-13
Prune-Depth Config Documentation, Added Unit Tests for Different Limits

nairashwin952013:
Configuration Parameters for ```prune_empty_parent_depth```, which are functionally simple to describe, seem to be really difficult to communicate about when specific cases are concerned. This just emphasizes the need to make certain documentation changes and adding a few unit tests for better interpretation of the setting.


nairashwin952013:
This PR has been squashed and merged.

PR #496, opened 2019-03-07
Fixing Issue#480

### LINKED ISSUE #480: Integration tests should be structured better to support accounts that have SnapLinks disabled
When running just MantaClientIT, the following tests fail due to snaplinks being disabled:
```
MantaClientIT.canMoveDirectoryWithContents
MantaClientIT.canMoveFileToDifferentPrecreatedDirectory
MantaClientIT.canMoveFileToDifferentUncreatedDirectoryCreationEnabled
MantaClientIT.testHead
MantaClientIT.testPutJsonLink
MantaClientIT.testPutLink
```

We should probably have a separate integration test class that does only operations related to snaplinks and remove all tests that use snaplinks from MantaClientIT.

nairashwin952013:
This Integration should close #480.

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
@@ -29,7 +29,7 @@ import static com.joyent.manta.client.MantaClient.SEPARATOR;
  * <p>
  * Since we want to make it possible to run this test without a code change, this test throws a {@link
  * org.testng.SkipException} if snaplinks are disabled for the "manta.user" account, the method
- * {@link MantaClientSnapLinksIT#skipSnapLinkTest(String, String)} will skip this integration-test from running.
+ * {@link MantaClientSnapLinksIT#putSnapLinkAndSkipIfUnsupported(String, String)} will skip this integration-test from running.
  * </p>
  * <p>Remember to also pass system properties for client configuration (manta.user/etc.) or set the values in the
  * environment (MANTA_USER/etc).
@@ -65,6 +65,7 @@ public class MantaClientSnapLinksIT {
         IntegrationTestConfigContext.cleanupTestDirectory(mantaClient, testPathPrefix);
     }
 
+    @Test(dependsOnMethods = "testPutLink")
     public final void canMoveDirectoryWithContents() throws IOException {
         final String name = "source-" + UUID.randomUUID().toString();
         final String source = testPathPrefix + name + MantaClient.SEPARATOR;
@@ -72,6 +73,7 @@ public class MantaClientSnapLinksIT {
         moveDirectoryWithContents(source, destination);
     }
 
+    @Test(dependsOnMethods = "testPutLink")
     public final void canMoveFileToDifferentPrecreatedDirectory() throws IOException {
         final String name = UUID.randomUUID().toString();
         final String path = testPathPrefix + name;
@@ -85,6 +87,7 @@ public class MantaClientSnapLinksIT {
         Assert.assertEquals(movedContent, TEST_DATA);
     }
 
+    @Test(dependsOnMethods = "testPutLink")
     public final void canMoveFileToDifferentUncreatedDirectoryCreationEnabled() throws IOException {
         final String name = UUID.randomUUID().toString();
         final String path = testPathPrefix + name;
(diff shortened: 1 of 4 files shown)


java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private static final String TEST_DATA = "Arise,Awake And Do Not Stop Until Your Goal Is Reached.";
+
+    private MantaClient mantaClient;
+
+    private String testPathPrefix;
+
+    @BeforeClass
+    @Parameters({"usingEncryption"})
+    public void beforeClass(@Optional Boolean usingEncryption) throws IOException {
```
dekobon: Perhaps attempt to create a snaplink in the before method. If it fails with the message that snaplinks aren't supported, then we should throw a `SkipException` to indicate that the whole test class can't be run. This approach may have a problem, if so - I think the alternative is to throw the `SkipException` in each individual test.

nairashwin952013: Noted.

nairashwin952013: I am opting for the latter since I agree with you when you posited: *It will be relatively safer from the design perspective*. I have changed occurrences of ```putSnapLink()``` in the there tests that it was implemented in and didn't make any changes to the others, would you like me to include them in all ?

nairashwin952013: A final change will also be added to the class description which will state the conditions under which this ```integration-test``` will be skipped.

nairashwin952013: New changes passed in ```staging```.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
```diff
@@ -0,0 +1,225 @@
...
+        mantaClient.putDirectory(testPathPrefix, true);
+    }
+
+    @AfterClass
+    public void afterClass() throws IOException {
+        IntegrationTestConfigContext.cleanupTestDirectory(mantaClient, testPathPrefix);
+    }
+
+    @Test
+    public final void canMoveDirectoryWithContents() throws IOException {
```
dekobon: You can omit the `@Test` annotation if you have specified it at the class level.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
```diff
@@ -0,0 +1,225 @@
+/*
```
dekobon: The following tests still fail when snaplinks are disabled:

```
MantaClientSnapLinksIT.canMoveDirectoryWithContents
MantaClientSnapLinksIT.canMoveFileToDifferentPrecreatedDirectory
MantaClientSnapLinksIT.canMoveFileToDifferentUncreatedDirectoryCreationEnabled
```

They are not properly skipped.

nairashwin952013: Is there like a flag to check whether snaplinks are disabled for an account ?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
```diff
@@ -0,0 +1,219 @@
...
+
+        boolean sourceIsDeleted = !mantaClient.existsAndIsAccessible(source);
+        Assert.assertTrue(sourceIsDeleted, "Source directory didn't get deleted: "
+                + source);
+    }
+
+    private void skipSnapLinkTest(final String linkPath, final String objectPath) throws IOException, MantaClientHttpResponseException {
+        try {
+            mantaClient.putSnapLink( linkPath, objectPath, null);
+        } catch (MantaClientHttpResponseException e) {
```
nairashwin952013:
```
catch (MantaClientHttpResponseException e) {
          if (e.getServerCode().equals(SNAPLINKS_DISABLED_ERROR)) {
                final String message =
                        "This integration-test class can't be run since SnapLinks" +
                                "have been disabled for this account";
                LOG.warn(message);
                throw new SkipException(message, e);
            }
         throw e;
    }
```

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSnapLinksIT.java
nairashwin952013: Noted.

nairashwin952013:
Addressed all feedback provided, #496 is eligible to be merged into the ```release-branch```.

nairashwin952013:
Carol notified for her feedback.

nairashwin952013:
This Pull-Request has been squashed and merged onto the ```release-branch```.

Issue #479, opened 2019-02-08
Review correctness of integer division in floating point context


dekobon:
See: 

[AesCbcCipherDetails.java:124](https://github.com/joyent/java-manta/blob/master/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java#L124)

nairashwin952013:
This was an extremely fun bug to verify since it's nature was mathematical, I also got a chance to use the [Wolfram-Alpha-Calculator](https://www.wolframalpha.com/input/?i=floor(0.0625) which is a fun-tool like ```MATLAB``` to perform complex mathematical operations. To surmise, ```Math.floor()``` works perfectly fine since its returns the *lesser than closest integer towards +infinity* which is essentially required to calculate the ```blocks``` for ```contentLength``` in ```AESCBCCiphers``` while running [AesCbcCipherDetailsTest#calculateContentLengthWorks()](https://github.com/joyent/java-manta/blob/master/java-manta-client-unshaded/src/test/java/com/joyent/manta/client/crypto/AesCbcCipherDetailsTest.java#L98).
I understand why you reported this since the aforementioned function is a bit generic nevertheless we can safely **close** Joyent/java-manta#479.

nairashwin952013:
Although, I suggest I should come up with a pull-request to define the test-group:```unlimited-crypto``` in the ```testing-it``` file.

dekobon:
Ok, I just took a look at this. This is interesting:
```java
    public static void main(String[] argv) {
        int a = -902342342;
        int b = 3233337;
        int intDivision = a / b;
        double floor = Math.floor(a / b);
        double floorDiv = Math.floorDiv(a, b);
        double casted = (double)a / (double)b;
        double floorCasted = Math.floor(casted);

        System.out.printf("a / b as int = %d\n", intDivision);
        System.out.printf("floor(a / b) as double = %f\n", floor);
        System.out.printf("floorDiv(a / b) as double = %f\n", floorDiv);
        System.out.printf("floor((double)a / (double)b) as double = %f\n", floorCasted);
    }
```

Output:
```
a / b as int = -279
floor(a / b) as double = -279.000000
floorDiv(a / b) as double = -280.000000
floor((double)a / (double)b) as double = -280.000000
```
All three values are the same if `a` and `b` are positive. In our case, `Math.floorDiv()` might make the most sense because it converts the ints to a double and does the floor operation all at once.

What are your thoughts?

[...]

PR #497, opened 2019-03-08
Fix Issue 479

nairashwin952013:
Following Up After Investigation of Fixing #479.

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaObjectOutputStream.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaObjectOutputStream.java
@@ -325,7 +325,7 @@ public class MantaObjectOutputStream extends OutputStream {
      * <p>NOTE: This close method calls flush() on the inner {@link OutputStream}
      * and then closes it if possible.</p>
      *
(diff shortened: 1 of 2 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/client/crypto/AesCbcCipherDetails.java
```diff
@@ -121,8 +122,8 @@ static long calculateContentLength(final long plaintextSize, final int blockByte
 
         // CBC requires an IV an extra block for chaining to work
         if (!hasIV && plaintextSize >= blockBytes) {
-            final double blocks = Math.floor(plaintextSize / ((long) blockBytes));
-            calculatedContentLength = (((long)blocks + 1) * blockBytes);
+            final long blocks = Math.floorDiv(plaintextSize, ((long) blockBytes));
```
dekobon: Let's make this a separate PR and commit. This is the type of change that you want it to be highly visible.

nairashwin952013: Noted.

PR #474, opened 2019-01-30
Fix 470 and Update MantaErrorCode Enum Values

nairashwin952013:
Certain Inconsistencies were observed while fixing [Issue#470](https://github.com/joyent/java-manta/issues/470). Follow-Up:

These discrepancies were removed and subsequently enum values corrected for: 
- ```409:```[ENTITY_EXISTS_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L168). 
- ``` 400:```[INVALID_PATH_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L304). 
- ```409:```[JOB_STATE_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L345). 
- ```400:```[LINK_REQUIRED_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L398).  
- ```400:```[MAX_CONTENT_LENGTH_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L420). 
- ```409:```[MULTIPART_UPLOAD_CREATE_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L450).

Additional Values Edited by referring [source file](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js). 

Enum Values added for:
- ```401:```[INVALID_ALGORITHM_ERROR](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L198). 
- ```503:```[SHARDS_EXHAUSTED_ERROR*](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js#L408). 

Other Values Added by referring [source file](https://github.com/joyent/manta-muskie/blob/master/lib/errors.js).

PR #501, opened 2019-03-13
Fix 470 and Update MantaErrorCode Enum Values

nairashwin952013:
Following Up after #474 has to be closed prematurely due to ```dependency-vulnerabilities```.

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015-2017, Joyent, Inc. All rights reserved.
+ * Copyright (c) 2015-2019, Joyent, Inc. All rights reserved.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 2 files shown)


nairashwin952013:
This created merge conflicts will be solved when this branch is eligible for a merge onto the ```release-branch```.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -149,10 +616,11 @@ public String getCode() {
      * @param serverCode code value to search for. Null is acceptable
      * @return Manta error code enum associated with serverCode parameter
      */
+    @SuppressWarnings("Duplicates")
     public static MantaErrorCode valueOfCode(final String serverCode) {
```
dekobon: How do you plan on dealing with codes that are duplicated but not deprecated like `AuthorizationFailed`?

I went through the codes and found the following duplicated:
```
AuthorizationFailed
AuthorizationRequired
AuthorizationSchemeNotAllowed
ContentLengthRequired
ContentMD5Mismatch
DirectoryLimitExceeded
InvalidArgumentError
InvalidHttpAuthenticationToken
InvalidJobState
InvalidMaxContentLength
InvalidQueryStringAuthentication
InvalidResource
LocationRequired
MaxContentLengthExceeded
MultipartUploadInvalidArgument
OperationNotAllowedOnDirectory
OperationNotAllowedOnRootDirectory
SecureTransportRequired
SourceObjectNotFound
```

I might have missed some, so please double check.

nairashwin952013: - Noted, I was working on it as well. 
- I just pushed a [commit](https://github.com/joyent/java-manta/pull/501/commits/a9e095ad0143b975b60c61239431b641c7c90cc9) to tag all the duplicates that are present and aren't **deprecated**. 
- I will come up with a safe implementation soon to handle duplicates.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -127,6 +606,14 @@
     private static final AtomicReference<Map<String, MantaErrorCode>> LOOKUP =
```
dekobon: I suggest that you change from lazy instantiation here to instantiation explicitly within the static scope.

nairashwin952013: Done, Kindly Review the new changes made.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -180,4 +532,52 @@ public static MantaErrorCode valueOfCode(final Object object) {
...
+        return nonDeprecatedList;
+    }
+
+    /**
+     * Populates an Inclusive Error Map for {@link MantaErrorCode} by its code value.
+     *
+     * @return a thread-safe Manta Errors HashMap.
+     * @throws MantaClientException if a thread-safe access is not granted for Error Map.
+     */
+    public static AtomicReference<Map<String, MantaErrorCode>> createErrorMap() throws MantaClientException {
```
dekobon: You don't need to use `AtomicReference` when you instantiate in the static scope eagerly.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -180,4 +532,52 @@ public static MantaErrorCode valueOfCode(final Object object) {
...
         return valueOfCode(object.toString());
     }
+
+    /**
+     * Populates a Inclusive ArrayList for {@link MantaErrorCode} by its code value.
+     *
+     * @return a thread-safe Manta Errors List.
+     * @throws MantaClientException thrown if deprecated fields are not present or due to security violation.
+     */
+    public static List<MantaErrorCode> getNonDeprecatedList() throws MantaClientException {
```
dekobon: Is this still needed? Also why is it a `List` and not a `Set`?

nairashwin952013: I thought a ```List``` would be preferable since it allows duplicated elements while ```Set``` usually contains an unordered collection of distinct elements. Also traversing a Set can only be done in a forward direction whereas ```Lists``` are flexible(```both forward and backward```) in that regard.

dekobon: Don't we want all values to be unique? Ordering for `Set` is an optional concern. There are several implementations of `java.util.NavigableSet` that allow for ordered sets that provide means for bidirectional traversal.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
nairashwin952013: Noted.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
nairashwin952013: ```List``` Removed.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -180,4 +522,31 @@ public static MantaErrorCode valueOfCode(final Object object) {
...
     }
+
+    /**
+     * Populates an Inclusive Error Map for {@link MantaErrorCode} by its code value.
+     *
+     * @return a thread-safe Manta Errors HashMap.
+     * @throws MantaClientException if a thread-safe access is not granted for Error Map.
+     */
+    public static Map<String, MantaErrorCode> createErrorMap() throws MantaClientException {
+        errorMap = Collections.synchronizedMap(new HashMap<>(values().length));
```
dekobon: Why are you assigning the map here in the method that is returning a reference to be assigned? Please be careful not to do this type of operation in the future - there is almost never a good case for doing this.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -180,4 +522,31 @@ public static MantaErrorCode valueOfCode(final Object object) {
...
+        for (MantaErrorCode m : values()) {
+            try {
+                Field field = MantaErrorCode.class.getField(m.name());
```
dekobon: Since we no longer have duplicated error codes, we no longer need to skip loading deprecated values. Something simple that validates the contract not to duplicate error codes like the following should be fine:
```java
    public static Map<String, MantaErrorCode> createErrorMap() throws MantaClientException {
        final Map<String, MantaErrorCode> map = new LinkedHashMap<>(values().length);

        for (MantaErrorCode errorCode : values()) {
            final MantaErrorCode replaced = map.put(errorCode.code, errorCode);

            if (replaced != null) {
                String msg = String.format("Duplicate error code specified in enum: %s",
                        replaced.code);
                throw new IllegalArgumentException(msg);
            }
        }

        return map;
    }
```

nairashwin952013: Noted.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
nairashwin952013: Noted.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
```diff
@@ -180,4 +520,26 @@ public static MantaErrorCode valueOfCode(final Object object) {
 
         return valueOfCode(object.toString());
     }
+
+    /**
+     * Populates an Inclusive Error Map for {@link MantaErrorCode} by its code value.
+     *
+     * @return a thread-safe Manta Errors HashMap.
```
dekobon: This is no longer thread-safe.

java-manta-client-unshaded/src/test/java/com/joyent/manta/exception/MantaErrorCodeTest.java
```diff
@@ -14,14 +14,22 @@
...
  *
  * @author <a href="https://github.com/dekobon">Elijah Zupancic</a>
+ * @author <a href="https://github.com/nairashwin952013">Ashwin A Nair</a>
  */
 @Test
 public class MantaErrorCodeTest {
     public void valueOfCodeCanFindByMatchingCode() {
-        final MantaErrorCode expected = MantaErrorCode.INVALID_LIMIT_ERROR;
-        final MantaErrorCode actual = MantaErrorCode.valueOfCode("InvalidLimit");
+        final MantaErrorCode expectedLimitError = MantaErrorCode.INVALID_LIMIT_ERROR;
```
dekobon: I would just use the raw enum because that is simpler.

java-manta-client-unshaded/src/main/java/com/joyent/manta/exception/MantaErrorCode.java
nairashwin952013: Noted.

java-manta-client-unshaded/src/test/java/com/joyent/manta/exception/MantaErrorCodeTest.java
nairashwin952013: Imported ```raw enums``` **statically** that were used in the added unit tests and the existing ones.

nairashwin952013:
This branch was successfully merged onto the ```release-branch``` PR #503  and all merge-conflicts were successfully resolved preserving the changes that were needed.

PR #492, opened 2019-03-04
Fixing Issue 483 For Defining HeavyWeight Test Groups

nairashwin952013:

PR #504, opened 2019-03-25
Defining HeavyWeight and Other Test Groups

nairashwin952013:
# Overview

Adds a `it.dryRun` property that can be used to list the tests that would run with their respective TestNG params. TestNG unfortunately doesn't completely support filtering a test suite by groups so thorough detailing was required in suite definition.

Output of `mvn verify it.dryRun=true` using current `testng-it.xml` used with `TestListingInterceptor`:
```
[main] INFO  com.joyent.test.util.MantaPathSuiteListener [ ] - Base manta path for suite Java Manta SDK Integration Test Suite: ~~/stor/java-manta-integration-tests/-some UUID-

[main] INFO  com.joyent.test.util.MantaPathSuiteListener [ ] - Expected test count: 229

[main] WARN  com.joyent.test.util.MantaPathSuiteListener [ ] - Skipping suite cleanup since dry-run is enabled.

[main] INFO  com.joyent.test.util.TestListingInterceptor [ ] - DRY-RUN: Listing [229] tests that would have run

```

List for the output for [before](https://gist.github.com/tjcelaya/fa82e54bd4be55ec95e1146954d7bd1f) & [after](https://gist.github.com/nairashwin952013/ae0a4799525654588bcce24e9595fd6a)  `-Dit.dryRun=true` respectively . Note that`dry-run-after` list is accurate since it shows single value for `encryptionCipher` i.e `AES128/CTR/NoPadding`. Other testing groups include `cbc-cipher` and `gcm-cipher` are also defined. 

# Changes

- change `usingEncryption` (boolean) to just `encryptionCipher` (string). omission (null) disables encryption

- add a new constructor for `IntegrationTestConfigContext` which handles the above

- switch from `@BeforeClass` with `@Parameters` to test constructors with just `@Parameters`. Whenever `SkipException` was thrown or `putDirectory` was called in a test which needed `@Parameters` a minimal method was created for `@BeforeClass`. The goal here was to make it safe to instantiate test classes without incurring side-effects.

- remove encryption param from tests which don't actually perform any object PUTs or MPU uploads (e.g. `MantaClientDirectoriesIT`)

- change `<packages>` to `<classes>` and target specific classes. 

- document a new system property `it.dryrun` which can be used by the new interceptor. When this system property is truthy as defined by `BooleanUtils#toBooleanObject`, i.e. the following are all true: `y`, `Y`, `t`, `T`, `on`, `ON`, `yes`, `YES`, `true`, `TRUE` and all the mixed-case versions thereof

- added a test listener called `TestListingInterceptor` which is used to collect suite definitions and disable actually running tests in the event `it.dryrun` is `true`.

- defined new `testng-groups` and the **EXISTING** ones  in `testing-it.xml` more accurately using class-level specifications including `jobs`, `snaplinks`, `nightly`, `seekable`, `mtime`, `error`, `headers`, `directory`, `metadata`, `move`

- Following up on #492 which **had** to be closed due to ```dependency-vulnerabilities```.

- Changes are incorporated from the closed PR #364 by Tomas which is to remove redundancies in ```integration-test``` runs.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -64,6 +64,17 @@ possible torun a subset of the test suite by adding
 mvn test -DexcludedGroups=unlimited-crypto
 ```
 
+## Dry run
+	
+It's possible to print a list of which integration tests would run and their relevant
+parameters using the following invocation:
+	
+```
+mvn verify -Dit.dryrun=true
+```
+	
+Note: This has not yet been expanded to unit tests.
+	
 # Writing Tests
 
 The `java-manta-client` module contains unit tests. Integration tests generally
--- a/TESTING.md
+++ b/TESTING.md
@@ -66,8 +66,8 @@ mvn test -DexcludedGroups=unlimited-crypto
 
 ## Dry run
 	
-It's possible to print a list of which integration tests would run and their relevant
-parameters using the following invocation:
+The following invocation will print a list of which integration tests would run and their relevant
+parameters::
 	
 ```
 mvn verify -Dit.dryrun=true
--- a/TESTING.md
+++ b/TESTING.md
@@ -70,7 +70,7 @@ The following invocation will print a list of which integration tests would run 
 parameters::
 	
 ```
-mvn verify -Dit.dryrun=true
+mvn verify -Dit.dryRun=true
 ```
 	
 Note: This has not yet been expanded to unit tests.
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,0 +1,44 @@
+# Configuration Parameter Contributions
+
+This is a brief guide for developers interested in making open-source contributions to the configuration parameters for 
+java-manta. [Java-Manta]((http://joyent.github.com/java-manta)) is a community-maintained Java SDK for interacting with 
+Joyent's Manta object storage system.
+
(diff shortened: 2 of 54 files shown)


TESTING.md
```diff
@@ -64,6 +64,17 @@ possible torun a subset of the test suite by adding
 mvn test -DexcludedGroups=unlimited-crypto
 ```
 
+## Dry run
+	
+It's possible to print a list of which integration tests would run and their relevant
```
dekobon: This may be better phrased:

> The following invocation will print a list of which integration tests would run and their relevant
parameters:

nairashwin952013: Noted.

dekobon:
Ashwin, can you write up in this PR a big-picture description of what this set of changes is intending to accomplish and how?

nairashwin952013:
> Ashwin, can you write up in this PR a big-picture description of what this set of changes is intending to accomplish and how?

Done.

nairashwin952013:
Test passed in ```staging```.

nairashwin952013:
This Pull request is closed due to `control-version` checks and other issues and will have a `duplicate-PR` in #507.

Issue #525, opened 2019-05-21
MantaClientSigningIT tests fail with MANTA_TLS_INSECURE=1


cburroughs:
`MantaClientSigningIT` tests fail with `MANTA_TLS_INSECURE=1`.  This is because they use their own `java.net.HttpURLConnection` since they are fetching the signed url *from* -- not *through* -- the MantaClient.

 I'm not sure how I missed this for #509 :-(

Example:
```
[ERROR] testCanCreateSignedURIWithEncodedCharacters(com.joyent.manta.client.MantaClientSigningIT)  Time elapsed: 0.115 s  <<< FAILURE!
javax.net.ssl.SSLHandshakeException: java.security.cert.CertificateException: No subject alternative names present
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
	at sun.security.ssl.SSLSocketImpl.fatal(SSLSocketImpl.java:1946)
	at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:316)
	at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:310)
	at sun.security.ssl.ClientHandshaker.serverCertificate(ClientHandshaker.java:1639)
	at sun.security.ssl.ClientHandshaker.processMessage(ClientHandshaker.java:223)
	at sun.security.ssl.Handshaker.processLoop(Handshaker.java:1037)
	at sun.security.ssl.Handshaker.process_record(Handshaker.java:965)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1064)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1367)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1395)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1379)
	at sun.net.www.protocol.https.HttpsClient.afterConnect(HttpsClient.java:559)
	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:185)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1564)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1492)
[...]

PR #526, opened 2019-05-21
allow signed url integration tests to work with MANTA_TLS_INSECURE

cburroughs:
ref #525

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015-2017, Joyent, Inc. All rights reserved.
+ * Copyright (c) 2015-2019, Joyent, Inc. All rights reserved.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -23,12 +23,20 @@ import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.nio.charset.Charset;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSession;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
 
 import static com.joyent.manta.client.MantaClient.SEPARATOR;
 import static java.nio.charset.StandardCharsets.UTF_8;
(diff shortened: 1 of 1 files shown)


java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -23,12 +23,21 @@
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.nio.charset.Charset;
+import java.security.SecureRandom;
+import java.security.SecureRandom;
```
nairashwin952013: Redundant Import

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -65,6 +74,51 @@ public void afterClass() throws IOException {
...
+                    @Override
+                    public void checkClientTrusted(X509Certificate[] certs, String authType) {
+                    }
+                    @Override
+                    public void checkServerTrusted(X509Certificate[] certs, String authType) {
+                    }
+                }
+            };
+
+             HostnameVerifier hostnameVerifier = new HostnameVerifier() {
```
nairashwin952013: This code block could be replaced with a lambda

```
HostnameVerifier hostnameVerifier = new HostnameVerifier() {
                @Override
                public boolean verify(String hostname, SSLSession session) {
                    return true;
                }
            };
```
to 

`HostnameVerifier hostnameVerifier = (hostname, session) -> true;`

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -65,6 +74,51 @@ public void afterClass() throws IOException {
...
+                @Override
+                public boolean verify(String hostname, SSLSession session) {
+                    return true;
+                }
+            };
+
+            try {
+                SSLContext sc = SSLContext.getInstance("SSL");
```
nairashwin952013: apologies for the annoying nit but could we *refactor* `sc` & `sConn` to more readable variables.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -65,6 +74,51 @@ public void afterClass() throws IOException {
...
     }
 
+
+    // When MantaClient TLS security is disabled (such as when testing against a
+    // lab instance without a self signed cert), we will also need to refrain
+    // from checking the certs on the resulting signed urls.
+    private HttpURLConnection openHttpURLConnection(URI uri, MantaClient client) throws IOException {
+        boolean tlsInsecure = client.getContext().tlsInsecure();
+        HttpURLConnection connection = (HttpsURLConnection)uri.toURL().openConnection();
+        if (tlsInsecure && connection instanceof HttpsURLConnection) {
```
nairashwin952013: Doubt: Could we replace `instanceof` condition with a `null` check ?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -65,6 +74,51 @@ public void afterClass() throws IOException {
         IntegrationTestConfigContext.cleanupTestDirectory(mantaClient, testPathPrefix);
     }
 
+
+    // When MantaClient TLS security is disabled (such as when testing against a
```
nairashwin952013: Comment could be replaced as a `javadoc`.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
cburroughs: The `instanceof` is for checking to make sure it is an `HttpsURLConnection` (subclass of `HttpURLConnection`).  I think almost all manta standups -- even lab ones -- use https, but if someone had an http only dev setup I didn't want to break that.

PR #528, opened 2019-05-23
Removing the usage of an extra MantaClient.SEPARATOR from the test cases

1010sachin:
The purpose of this PR is to remove extraneous usage of '/' in the path of the directories created during the execution of various test cases. This will provide uniformity in the test directory path creation and help to extend and write newer test cases in the future.

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientDirectoriesIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientDirectoriesIT.java
@@ -54,9 +54,8 @@ public class MantaClientDirectoriesIT {
     public void beforeClass() throws IOException {
         config = new IntegrationTestConfigContext();
         mantaClient = new MantaClient(config);
-        testPathPrefix = IntegrationTestConfigContext.generateBasePath(config, this.getClass().getSimpleName());
-        //Remove the extra '/' from the testPathPrefix
-        testPathPrefix = testPathPrefix.substring(0, testPathPrefix.lastIndexOf(SEPARATOR));
+        testPathPrefix = IntegrationTestConfigContext.generateBasePathWithoutSeparator(config,
+                this.getClass().getSimpleName());
     }
 
     @AfterClass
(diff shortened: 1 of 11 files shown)


dekobon:
Could you provide a better description of the change in the PR description?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientDirectoriesIT.java
```diff
@@ -55,6 +55,8 @@ public void beforeClass() throws IOException {
         config = new IntegrationTestConfigContext();
         mantaClient = new MantaClient(config);
         testPathPrefix = IntegrationTestConfigContext.generateBasePath(config, this.getClass().getSimpleName());
```
dekobon: Would it be possible to modify the `IntegrationTestConfigContext.generateBasePath` method rather than stripping the last character? Alternatively, another method called `IntegrationTestConfigContext.generateBasePathWithoutSeparator` (or some better name) that omits the seperator.

1010sachin: I did that originally, but that method is used at various places, for eg: ```MantaClientAuthenticationChangeIT``` and in most cases the '/' at the end of the testPathPrefix is required. So for such cases, we would have to add a SEPARATOR. So I left it there and did the stripping. But yeah, another method in the ```IntegrationTestConfigContext``` sounds good. Will add that change.

1010sachin: Added a new method and removed the code that strips the extra '/' from the testPathPrefix.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientErrorIT.java
```diff
@@ -24,6 +24,7 @@
 import java.util.Properties;
 import java.util.UUID;
 
+import static com.joyent.manta.client.MantaClient.SEPARATOR;
```
nairashwin952013: Unused `import`, Kindly remove it

nairashwin952013:
All dependent `integration-tests` pass.

nairashwin952013:
apart from the unused `import` in `MantaClientErrorIT`, LGTM!

PR #529, opened 2019-05-24
Update changelog to conform with local idiom/standard

russelldb:
I'd not looked enough at the surroundning entries to understand the
changelog format/expectations.

I have to ask, why this CHANGELOG.md when there is a git log?
And also, what is the protocol for updating the changelog? Is there some docs I can read as to what the sections mean and what I am supposed to write in them?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This project aims to adhere to [Semantic Versioning](http://semver.org/).
 
 ## [3.4.1]
 ### Added
- - [Fixed bug](https://github.com/joyent/java-manta/pull/527) where
+ - [Updated](https://github.com/joyent/java-manta/pull/527) where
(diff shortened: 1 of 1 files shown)


nairashwin952013:
`LGTM!`

nairashwin952013:
> I'd not looked enough at the surroundning entries to understand the
> changelog format/expectations.
> 
> I have to ask, why this CHANGELOG.md when there is a git log?
> And also, what is the protocol for updating the changelog? Is there some docs I can read as to what the sections mean and what I am supposed to write in them?

I think we have the `CHANGELOG.md` to track the progress of the SDK across different `version` iterations and documents the disparities between each successive version form its preceding counterpart. Mostly `ADDED` section is a new feature that has been incorporated into the `SDK`.

`FIXED` is for bugs that are fixed and `CHANGED` is for any existing feature in the `SDK` that has been. codified due to customer requirements/ design optimizations

PR #536, opened 2019-06-06
Add MANTA_BUCKETS_ENABLED Configuration Parameter

nairashwin952013:
## Goals
- Added a new `config` parameter for Manta Buckets Usage in `JAVA SDK`.

- Setting `manta.buckets=true` \ `MANTA_BUCKETS=true` as an `env` variable will enable the client to use bucket operations enumerated as:
              - `createBucket`
              - `deleteBucket`
              - `listBucket` 
and so on.

- Naming convention for the said `config` parameter is chosen with the intent to make it consistent with `enableBuckets` evident throughout the different components of manta including muskie, electric-boray, boray etc.

- This configuration parameter is also vital to running `integration-tests` for buckets, since the base path for running buckets will be `/$MANTA_USER/buckets/`.

- DEFAULT_BUCKETS_ENABLED has been set to `false` since we don't want to have directory operations(API calls) conflicting with the Buckets API Calls.

## Pending
- More information in Usage will be added once we have [buckets-operations](https://github.com/nairashwin952013/java-manta/tree/buckets-operations) merged into the base-branch: `buckets-m1`.

## Implementation
- Steps to add a `config` parameter to the Java SDK can be found in [DEVELOPING.md](https://github.com/joyent/java-manta/blob/master/DEVELOPING.md).

## Correctness
- Changes have also been made to ensure that the configuration parameter value when supplied in `context` will override the defaults.


dekobon:
I believe that the approach specified in this PR is not a good approach. Buckets is a feature of the server-side API and should not require being enabled in the SDK. Adding this increases the burden placed upon the users of the SDK and doesn't provide any apparent benefit.

If users attempt to use buckets features against a Manta install that does not support them, then the SDK should error and ideally inform them that the Manta install doesn't support buckets.

nairashwin952013:
Based on 

> I believe that the approach specified in this PR is not a good approach. Buckets is a feature of the server-side API and should not require being enabled in the SDK. Adding this increases the burden placed upon the users of the SDK and doesn't provide any apparent benefit.
> 
> If users attempt to use buckets features against a Manta install that does not support them, then the SDK should error and ideally inform them that the Manta install doesn't support buckets.

That makes sense, we should probably just configure the existing integration tests in such a way that it works both for `buckets` and `directories`.

I am gonna go ahead and close PR #536

PR #539, opened 2019-06-10
Add a new IntegrationTestHelper class for buckets.

1010sachin:
### Goal
The goal of this PR is to provide a test helper class that can be use to extend the current Integration Test suite to test buckets api, when they are ready.

### Implementation
* Introduce a new TestNG parameter, called ```String testType``` that when set to the value ```buckets``` will setup up the ```bucketsTestPathPrefix``` to the required bucket format i.e. ```/$MANTA_USER/buckets/bucket_name``` and the ```testPathPrefix``` to ```/$MANTA_USER/buckets/bucket_name/objects/object_name```.
* Introduce a new helper class, called ```IntegrationTestHelper.java``` that contains utility methods to verify whether the MANTA server supports the buckets API, configure the paths to test the buckets and bucket objects API in the test classes and clean up the bucket and objects after execution of the test cases.

*NOTE This PR contains changes in the ```MantaClientIT.java``` to show the use of the class ```IntegrationTestHelper```. The same changes can be propagated to the other classes when the buckets API code is ready for testing.*


java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -57,25 +60,48 @@
...
 
         // Let TestNG configuration take precedence over environment variables
         ConfigContext config = new IntegrationTestConfigContext(usingEncryption);
+        String testName = this.getClass().getSimpleName();
 
         mantaClient = new MantaClient(config);
-        testPathPrefix = IntegrationTestConfigContext.generateBasePath(config, this.getClass().getSimpleName());
-        mantaClient.putDirectory(testPathPrefix, true);
+
+        if(testType != null && testType.equals("buckets")) {
```
dekobon: Please put a space after `if`.

dekobon: You can make this conditional simpler by doing:

```java
if ("buckets".equals(testType))
```

This way if testType is `null`, it will evaluate to false and not throw a NPE.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -57,25 +60,48 @@
...
+            if(IntegrationTestHelper.verifyBucketsSupport(config, mantaClient)) {
```
dekobon: Please add a space after `if`.

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -57,25 +60,48 @@
...
+                bucketsEnabled = true;
```
dekobon: Could all of this logic be put in a helper? Does it have to be duplicated in every integration test?

java-manta-it/src/test/java/com/joyent/manta/client/MantaClientIT.java
```diff
@@ -57,25 +60,48 @@
...
+            testPathPrefix = IntegrationTestConfigContext.generateBasePath(config, testName);
+            mantaClient.putDirectory(testPathPrefix, true);
+        }
+
     }
 
     @AfterClass
     public void afterClass() throws IOException {
-        IntegrationTestConfigContext.cleanupTestDirectory(mantaClient, testPathPrefix);
+        if(BooleanUtils.isTrue(bucketsEnabled)) {
```
dekobon: Could the `IntegrationTestHelper` have a more generic method that would do the right thing based on the `testType`? Is there a way in which we could lessen duplication?

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+ * A helper class that contains methods to assist in the Integration testing of
+ * manta buckets functionality.
+ *
+ * @author <a href="https://github.com/1010sachin">Sachin Gupta</a>
+ */
+
+public class IntegrationTestHelper {
+
+    public static String setupBucketsPath(final ConfigContext config, final String testName) {
+        String testPathPrefix = String.format("%s/buckets/%s", config.getMantaHomeDirectory(),
```
dekobon: Please use `MantaClient.SEPERATOR` and do not hardcode slashes.

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+                testName + UUID.randomUUID());
+        return testPathPrefix + MantaClient.SEPARATOR;
+    }
+
+    public static boolean verifyBucketsSupport(final ConfigContext config,
+                                               final MantaClient client) {
+        AtomicBoolean isBucketsSupported = new AtomicBoolean();
```
dekobon: `AtomicBoolean` is unneeded here. Please research the correct usage of it. A normal boolean variable is fine.

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+        try {
+            String path = String.format("%s/buckets", config.getMantaHomeDirectory());
```
dekobon: Use `MantaClient.SEPARATOR`.

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+            client.get(path);
```
dekobon: Let's wait until @kellymclaughlin commits an `OPTIONS` call before we merge this PR, so that we can do the proper call to detect if buckets are supported.

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+            isBucketsSupported.set(true);
+        } catch (IOException io) {
```
dekobon: Don't just catch the generic `IOException` because it could be thrown for many different problems beyond the case you are looking for. You may need to do more precise matching on the error message or a more precise exception type.

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
```diff
@@ -0,0 +1,77 @@
...
+                                       final String testType) {
+        if ("buckets".equals(testType)) {
+            if (BooleanUtils.isTrue(verifyBucketsSupport(config, client))) {
+                return setupBucketsPath(config, testName);
+            } else {
+                throw new SkipException("Bucket operations not supported by the server");
+            }
+        }
+
+        return IntegrationTestConfigContext.generateBasePath(config, testName);
```
nairashwin952013: We want to make sure that the bucket name doesn't start with an upper case character, since it's one of the naming restrictions as mentioned [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html).

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
nairashwin952013: This generic exception could be replaced with:
```
catch (MantaClientHttpResponseException e) {
            if (e.getServerCode().equals(RESOURCE_NOT_FOUND_ERROR)) {
               isBucketsSupported.set(false);
            }
```

nairashwin952013:
Merely a suggestion but you could update the copyright headers for all the changed files.

java-manta-it/src/test/java/com/joyent/manta/client/helper/IntegrationTestHelper.java
1010sachin: Okay.

PR #559, opened 2019-12-09
Dependent test modifications to verify MANTA-4839

nairashwin952013:
Introduce some test modifications to comply with Manta `v1/v2` changes. Refer: [MANTA-4839](https://jira.joyent.us/browse/MANTA-4839)

### DIFF
--- a/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
+++ b/java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
@@ -10,6 +10,7 @@ package com.joyent.manta.client;
 import com.joyent.manta.config.ConfigContext;
 import com.joyent.manta.config.IntegrationTestConfigContext;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


java-manta-it/src/test/java/com/joyent/manta/client/MantaClientSigningIT.java
```diff
@@ -261,7 +263,7 @@ public final void testCanCreateSignedOPTIONSUriFromPath()
                 Assert.fail(errorText);
             }
 
-            String server = connection.getHeaderField("Server");
+            String server = StringUtils.substringBefore(connection.getHeaderField("Server"), SEPARATOR);
```
trentm: Tho this server header uses `/` and `com.joyent.manta.client.MantaClient.SEPARATOR` happens to also be `/`, I don't think this code needs to related them. I'd be inclined to use a literal `'/'` here. Does that work?

nairashwin952013: Feedback accounted for: refactored `SEPARATOR -> "/" literal`

PR #567, opened 2020-04-07
MANTA-5040 Dependency upgrade for java-client

nairashwin952013:
Dependency upgrade for all dependent libraries used by java-manta


nairashwin952013:
merging #567 would result in a version upgrade for java-manta from `3.4.3` -> `3.5.0`

PR #570, opened 2020-04-16
Fixes #468

### LINKED ISSUE #468: Implementation of MantaSeekableByteChannel.read(ByteBuffer) is broken
Looking at our implementation of [MantaSeekableByteChannel.read(ByteBuffer)](https://github.com/joyent/java-manta/blob/master/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java#L189) it appears that we don't follow the API contract:
```java
    public int read(final ByteBuffer dst) throws IOException {
        if (!open) {
            throw new ClosedChannelException();
        }

        final MantaObjectInputStream stream = connectOrGetResponse();
        final long size = size();

        if (position.get() >= size) {
            return EOF;
        }

        final byte[] buff = dst.array();
        final int bytesRead = stream.read(buff);

        position.addAndGet(bytesRead);

        return bytesRead;
    }
```
Interface documentation:
```java
    /**
     * Reads a sequence of bytes from this channel into the given buffer.
     *
     * <p> Bytes are read starting at this channel's current position, and
     * then the position is updated with the number of bytes actually read.
     * Otherwise this method behaves exactly as specified in the {@link
     * ReadableByteChannel} interface.
     */
```

This method is **not** reading data into the supplied byte buffer at all. In the Java source code, many of the implementations that implement this method use the static method [`IOUtil.read`](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/20e6cadfac43/src/share/classes/sun/nio/ch/IOUtil.java#l185). Within that implementation you can see that data is clearly being read into the `ByteBuffer`. Whereas in the `MantaSeekableByteChannel` implementation no data is being read into the `ByteBuffer`.

nairashwin952013:
Looking at our implementation of [MantaSeekableByteChannel.read(ByteBuffer)](https://github.com/joyent/java-manta/blob/master/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java#L189) it appears that we don't follow the API contract:

Interface documentation:
```java
    /**
     * Reads a sequence of bytes from this channel into the given buffer.
     *
     * <p> Bytes are read starting at this channel's current position, and
     * then the position is updated with the number of bytes actually read.
     * Otherwise this method behaves exactly as specified in the {@link
     * ReadableByteChannel} interface.
     */
```

This method is **not** reading data into the supplied byte buffer at all. In the Java source code, many of the implementations that implement this method use the static method [`IOUtil.read`](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/20e6cadfac43/src/share/classes/sun/nio/ch/IOUtil.java#l185). Within that implementation you can see that data is clearly being read into the `ByteBuffer`. Whereas in the `MantaSeekableByteChannel` implementation no data is being read into the `ByteBuffer`.

### DIFF
--- a/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
+++ b/java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
+ * Copyright (c) 2016-2019, Joyent, Inc. All rights reserved.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -30,7 +30,8 @@ import java.util.concurrent.atomic.AtomicReference;
  * Manta. Connection opening to the remote server happens lazily upon the
  * first read() or size() method invoked.
  *
- * @author Elijah Zupancic
+ * @author <a href="https://github.com/dekobon">Elijah Zupancic</a>
+ * @author <a href="https://github.com/nairashwin952013">Ashwin A Nair</a>
  */
 public class MantaSeekableByteChannel extends InputStream
         implements SeekableByteChannel {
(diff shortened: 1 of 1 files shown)


java-manta-client-unshaded/src/main/java/com/joyent/manta/client/MantaSeekableByteChannel.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017, Joyent, Inc. All rights reserved.
+ * Copyright (c) 2016-2019, Joyent, Inc. All rights reserved.
```
kellymclaughlin, line 2: 2020

